PR #11, opened 2021-10-29
Add initial draft spec for version ranges

pombredanne:
Here are some of the discussions that led to the creation of this specification:

- https://github.com/package-url/purl-spec/issues/66
- https://github.com/package-url/purl-spec/issues/84
- https://github.com/package-url/purl-spec/pull/93
- https://github.com/nexB/vulnerablecode/issues/119
- https://github.com/nexB/vulnerablecode/issues/140

Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>


VERSION-RANGE-SPEC.rst
```diff
@@ -0,0 +1,537 @@
...
+###############################
+
+https://github.com/npm/node-semver#ranges
+
+The node semver is very similar to this spec (this is also an OR of ANDs) but it
+has a few practical issues:
+
+- The space means "AND" and significant whitespace in a single string makes
+  normalization more complicated and may be a source of confusion or errors. The
+  explicit ampersand used an "AND" operator specified here improves clarity.
```
Hritik14, line 477: used as :)

Hritik14 (commented):
There are a few things that caught my eye

VERSION-RANGE-SPEC.rst
```diff
@@ -0,0 +1,537 @@
...
+
+To parse a version range specifier string:
+
+- Remove all spaces and tabs.
+- Start from left, and split once on colon ":".
+- The left hand side is the URI-scheme that must be lowercase.
+  - Verify that the URI-scheme value is ``vers``.
+- The right hand side is the specifier.
+
+- Split the specifier from left once on colon ":".
```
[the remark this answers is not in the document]
pombredanne: good catch! yes

VERSION-RANGE-SPEC.rst
```diff
@@ -0,0 +1,537 @@
...
+
+There is no (mostly) universal notation for version ranges and there is no
+universal way to compare two versions, even though the concepts that exist in
+most version range notations are similar.
+
+Each package type or ecosystem may define their own ranges notation and version
+comparison semantics for dependencies. And for security advisories, the lack of
+portable notation for vulnerable package version ranges means that these ranges
+may be either ambiguous or hard to compute and are typically replaced by
+complete enumerations of all versions, such as the NVD CPEs.
```
sbs2001, line 45: I don't think NVD CPEs provide complete enumerations.

VERSION-RANGE-SPEC.rst
```diff
@@ -0,0 +1,537 @@
...
+  - Gentoo https://wiki.gentoo.org/wiki/Version_specifier
+  - Alpine linux https://gitlab.alpinelinux.org/alpine/apk-tools/-/blob/master/src/version.c
+
+- Go modules https://golang.org/ref/mod#versions use semver versions with
+  specific version resolution algorithms.
+
+- Haskell Package Versioning Policy https://pvp.haskell.org/ provides a notation
+  similar to this specification based on a modified semver with extra notations
+  such as star and caret.
+
```
sbs2001, line 93: Please mention Arch linux. They have their own vercmp thingy.

VERSION-RANGE-SPEC.rst
```diff
@@ -0,0 +1,537 @@
...
+
+Note that there is a closely related problem: the way two versions are compared
```
sbs2001, line 95: Nit: Maybe breakdown the problems into a bullet list ? This sentence is hard to follow

VERSION-RANGE-SPEC.rst
```diff
@@ -0,0 +1,537 @@
...
+---------
+
+A solution to the many version range syntaxes is to design a new notation to
+unify them all with:
+
+- a mostly universal and minimalist notation to express the version range
+  notations from many different package types and ecosystems.
+
+- the package type-specific definitions to normalize existing range expressions
+  to this common notation .
```
sbs2001, line 114: Nit: extra space before the period

VERSION-RANGE-SPEC.rst
```diff
@@ -0,0 +1,537 @@
...
+has been selected because it is short, obviously about version and available
+for a future formal registration for this URI-scheme at the IANA registry.
+
+
+``<versioning-scheme>``
+------------------------
+
+The ``<versioning-scheme>`` (such as ``semver``,
+``debian``, etc.) determines:
+
```
stevespringett, line 164: To my understanding, many of the version range schemas are not standards, they are policies or conventions. What happens when those conventions change over time? For example, if in a year, Debian changes how they represent relationships between packages, will that alter the meaning of every `vers:debian` being used?

Additionally, what happens when one of the schemes adopts another scheme? For example, if Maven were to adopt semver in the future?

VERSION-RANGE-SPEC.rst
```diff
@@ -0,0 +1,537 @@
...
+  - Verify that the URI-scheme value is ``vers``.
+- The right hand side is the specifier.
+
+- Split the specifier from left once on colon ":".
+- The left hand side is the <versioning-scheme> that must be lowercase.
+- The right hand side is the constraints.
+
+- If the constraint contains a star "*", validate that it is equal to "*".
+  <version-constraint> is "*". Parsing is done and no further processing is
+  needed for this ``vers``.
```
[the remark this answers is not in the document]
pombredanne: That's a tool implementation issue. A tool may prefer to be strict and report some error, or be lenient and ignore silently the rest. Spec-wise, what is important is what is valid or not valid. That said let me add exactly this phrase.

VERSION-RANGE-SPEC.rst
```diff
@@ -0,0 +1,537 @@
...
+- "=": Version equality comparator. It is the default and implied if not
+  present and means that a version must be equal to the provided version.
+  For example: "=1.2.3". It must be omitted in the canonical representation.
+  Equality is based on the equality of two lower-cased and normalized version
+  strings and is not versioning scheme-specific.
+
+- "!=": Version exclusion or inequality comparator. This means a version must
+  not be equal to the provided version and this version must be excluded from
+  the range. For example: "!=1.2.3" means that version "1.2.3" is not part of
+  the range.
```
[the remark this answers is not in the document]
pombredanne: > Eg: "!=1.2.3, >1.0.0" implies versions should be "greater than 1.0.0 or not equal to 1.2.3" which will consist a set of every possible version.

yes, but you may want to write instead "!=1.2.3 & >1.0.0" which means what you intended: ""greater than 1.0.0 and not equal to 1.2.3" 

Would you have a an example of something that cannot be expressed with this notation?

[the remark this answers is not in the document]
pombredanne: Could it be easier to read a "vers" if this is in CNF, e.g., an AND of ORs?
.. possibly using a coma for AND (like is done in pypi) and a pipe for OR?

- AND is the default for pypi (there is no OR) with a coma; same for debian, rpm and rust: AND is the first thing, and is a coma
- same for node-semver and composer where AND is the default

With that said, the two use cases may be conflicting:

- for vulnerabilities, it is common to have a the same vulnerability affect multiple branches, such as httpd 1.x and 2.x. Here a version that's within any of these ranges would be vulnerable. e.g. any woudl mean an OR first.

- for dependencies, we are trying to resolve and pick one version and in the common case, all the constrainst would need to be satisfied. Some notations do not even support an OR.


So vulnerabilities, OR is first. Dependencies AND is first.

In order to adopt the most common notations in use, it may be better to use AND first as this would be more natural.
But this would mean that for vulnerabilities multiple "vers" might need to be provided, essentially one for each "branch"; it not seem completely crazy but I am not sure I like it.


For now, I think keeping an OR as a the default is best. But adapting the notation to have a comma mean "AND" may be easier since that's what a coma means elsewhere.

VERSION-RANGE-SPEC.rst
```diff
@@ -0,0 +1,537 @@
...
+
+    - If the version contains a percent "%" character, apply URL quoting rules
+      to unquote this string.
+
+    - Yield the comparator and version for this constraint
+
+  - Yield the accumulated list of (comparator and version) that must apply for
+    this constraint
+
+- Finally return the <versioning-scheme> and the list of <version-constraint>
```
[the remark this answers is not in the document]
pombredanne: I would rather keep the spec free of code snippets.  But to your point, a tool could use a list of lists to represent the constraints.... I am working a rev of univers in a branch to match this spec, and this comes out a rather natural way to do model the constraints.

VERSION-RANGE-SPEC.rst
pombredanne: thx

VERSION-RANGE-SPEC.rst
```diff
@@ -0,0 +1,537 @@
...
+
+https://nvd.nist.gov/vuln/vulnerability-detail-pages#divRange
+
+The NVD CPE Match String Range is a complex specification that goes well beyond
+version ranges and is used to match comprehensive configurations across multiple
+products and version ranges. The notation is to uses these fields in the CVE API
+or feeds under "configurations"::
+
+    "versionStartIncluding": "7.3.0",
+    "versionEndExcluding": "7.3.31",
```
pombredanne, line 461: While looking into https://nvd.nist.gov/vuln/data-feeds#cpeMatch there are also:
```
    "versionStartExcluding" : "9.0.0",
    "versionEndIncluding" : "9.0.46",
```
and concrete enumeration of versions updated daily.

VERSION-RANGE-SPEC.rst
pombredanne: You are right, this is a bit more complex indeed.
1. there are version ranges as part of the configuration https://github.com/nexB/univers/pull/11/files#diff-dda3c48de6da2f63bfa5902e41755b27a2f1d8c186a21dcc2641c42d4f5cfaebR450
2. and there are daily recomputed concrete ranges available as a feed at https://nvd.nist.gov/vuln/data-feeds#cpeMatch ... which look to me as complete enumerations.

This feed contains both ranges and the enumerations. I wonder if they compute these or add them by hand?

VERSION-RANGE-SPEC.rst
pombredanne: thx

VERSION-RANGE-SPEC.rst
pombredanne: @stevespringett hey!
Thank you ++ for the feedback. That's a good point. 

Even though there may not always be a spec or clear documentation there is always code in the package management tools that provide something IMHO as good as a spec.

That said, to your point we cannot trust that upstream ecosystems will not change their ways. 

A few thoughts:
1. If they do this in a non-compatible way, this will wreak havoc on their ecosystem. Not a reassuring thought, but at least this makes these events less likely. But they can happen. See Go for instance.
2. When I helped cleanup the linux kernel licensing using SPDX identifiers there was a concern if the SPDX were to change. The solution was simply to incorporate the few bits we depended on in the kernel process documentation thus making the doc self standing (and this was a good decision as some key license identifiers were deprecated mid-way of this work). To make it short, we can incorporate these external specs as needed to "freeze" them in this spec.
3. And if they change in the future in a non-compatible, they would essentially create a new versioning scheme, and we could also create a new identifier for this versioning scheme such as `vers:debian2`, or `vers:maven-semver`

Unrelated, this makes me think that  `vers:semver` is not a versioning scheme IMHO ... I need to update the text: `node-semver` is rather it as used in a `package.json`: e.g. the version are `semver` but the version ranges are node-specific.

VERSION-RANGE-SPEC.rst
```diff
@@ -0,0 +1,537 @@
...
+portable notation for vulnerable package version ranges means that these ranges
+may be either ambiguous or hard to compute and are typically replaced by
+complete enumerations of all versions, such as the NVD CPEs.
+
+Because of this, expressing and resolving a version range is often a complex, or
+error prone task.
+
+In particular the need for common notation for version has emerged based on the
+usage of Package URLs referencing vulnerable package version ranges such as in
+vulnerability databases like VulnerableCode.
```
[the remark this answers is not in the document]
pombredanne: done

VERSION-RANGE-SPEC.rst
pombredanne: Tell me if this more readable with the latest push.

pombredanne:
I have been playing more with the code to implement this and what transpires from this is that there seems to be no such thing as a "versioning scheme" that would be different from a `purl` package type. Even when I thought that `semver` may have been something that applies to several ecosystem, each seem to have their own small variations that may mean that most code would be shared but still each package type may define as overall scheme.... so I am inclined to reuse the `purl` type instead of the versioning scheme specified here.

VERSION-RANGE-SPEC.rst
pombredanne: > Unrelated, this makes me think that vers:semver is not a versioning scheme IMHO ... I need to update the text: node-semver is rather it as used in a package.json: e.g. the version are semver but the version ranges are node-specific.


I evolved this and now there is rather almost a 1-to-1 between a purl type and a vers scheme

pombredanne:
I am merging this now here to get closure and will start a new PR at https://github.com/package-url/purl-spec for this spec

VERSION-RANGE-SPEC.rst
jbmaillet: Just as another example of a convention that changed over time is the most widely used database in the world (probably billions of instances, would it be only on our phones, including the pre-smartphone area): **sqlite**.

[In 2015 it changed its versioning scheme to semver](https://www.sqlite.org/versionnumbers.html). And with a twist to make matter worst: the downloadable archives are not named sqlite-MAJOR.MINOR.PATCHLEVEL, but sqliteMAJORMINORPATCHLEVELWITHPADDINGZEROS. For example [3.37.0 is 3370000](https://www.sqlite.org/download.html). (It also uses its own CVS, Fossil, not git, but that's another concern.)

PR #27, opened 2022-01-13
Use SemverVersion in NginxVersionRange

Hritik14:
Fixes: https://github.com/nexB/univers/issues/24

Signed-off-by: Hritik Vijay <hritikxx8@gmail.com>

PR #28, opened 2022-01-13
Add tests to check from_string & from_spec equality

Hritik14:
This will avoid errors like https://github.com/nexB/univers/pull/27
popping up in future

Signed-off-by: Hritik Vijay <hritikxx8@gmail.com>

### DIFF
--- a/tests/test_version_range.py
+++ b/tests/test_version_range.py
@@ -5,11 +5,13 @@
 # Visit https://aboutcode.org and https://github.com/nexB/univers for support and download.
 
 from unittest import TestCase
+import pytest
 
 from univers.version_constraint import VersionConstraint
 from univers.version_range import GemVersionRange
 from univers.version_range import PypiVersionRange
 from univers.version_range import VersionRange
+from univers.version_range import RANGE_CLASS_BY_SCHEMES
 from univers.versions import PypiVersion
 from univers.versions import RubygemsVersion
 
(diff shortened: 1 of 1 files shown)


pombredanne (approved):
Thank you : LGTM!

tests/test_version_range.py
```diff
@@ -150,3 +152,19 @@ def test_GemVersionRange_from_native_range_with_pessimistic_operator(self):
             VersionConstraint(comparator=">=", version=RubygemsVersion(string="2.0.8")),
             VersionConstraint(comparator="<", version=RubygemsVersion(string="2.1")),
         )
+
+
+@pytest.mark.parametrize("scheme, range_class", RANGE_CLASS_BY_SCHEMES.items())
+class TestVersionRangeImplementations:
+    VERSION_RANGES_BY_SCHEMES = {
```
pombredanne, line 159: What about making this what you parametrize on?
And adding a separate, non-parametrized test that fails if there is not one test for every supported/known vers/purl types?

tests/test_version_range.py
```diff
@@ -150,3 +152,19 @@ def test_GemVersionRange_from_native_range_with_pessimistic_operator(self):
...
+        "nginx": ["0.8.40+", "0.7.52-0.8.39", "0.9.10", "1.5.0+, 1.4.1+"],
+    }
+
+    def test_from_native_and_from_string_equality(self, scheme, range_class):
+        if scheme not in self.VERSION_RANGES_BY_SCHEMES:
```
pombredanne, line 164: As mentioned above, you should not iterate and skip if not present. Instead iterate on the actual tests  and not the range class... and have a test that fails if we do not cover all the schemes.

Hritik14: But this test will almost certainly fail as we have many unimplemented VersionRange

pombredanne: not if marked as expected failure

Hritik14: Thanks!

Issue #30, opened 2022-01-14
Version in VerisonRange`"vers:nginx/*` crashes for every version


Hritik14:
```py
>>> SemverVersion("1.0.0") in VersionRange.from_string("vers:nginx/*")
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-12-b59bd155f0a0> in <module>
----> 1 SemverVersion("1.0.0") in VersionRange.from_string("vers:nginx/*")

~/Contrib/univers/src/univers/version_range.py in __contains__(self, version)
    159                 f"{version!r} is not of expected type: {self.version_class!r}",
    160             )
--> 161         return contains_version(version, self.constraints)
    162
    163     contains = __contains__

~/Contrib/univers/src/univers/version_constraint.py in contains_version(version, constraints)
    443     # Check is finished.
    444     if len(constraints) == 1:
--> 445         return version in constraints[0]
    446
    447     # If the "tested version" is equal to the any of the constraint version where

~/Contrib/univers/src/univers/version_constraint.py in __contains__(self, version)
    213
    214         if not isinstance(version, self.version.__class__):
--> 215             raise ValueError(
    216                 f"Cannot compare {version.__class__!r} instance "
    217                 f"with {self.version.__class__!r} instance."

ValueError: Cannot compare <class 'univers.versions.SemverVersion'> instance with <class 'NoneType'> instance.
```

PR #31, opened 2022-01-14
Fix __contains__ for vers:ANY/*

Hritik14:
Fixes: https://github.com/nexB/univers/issues/30

Signed-off-by: Hritik Vijay <hritikxx8@gmail.com>

### DIFF
--- a/src/univers/version_constraint.py
+++ b/src/univers/version_constraint.py
@@ -66,11 +66,14 @@ class VersionConstraint:
     # one of the COMPARATORS
     comparator = attr.ib(type=str, default="=")
 
-    # a Version subclass instance or None
+    # a Version subclass instance
     version = attr.ib(type=Version, default=None)
 
     # a function for the comparator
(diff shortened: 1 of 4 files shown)


src/univers/version_constraint.py
```diff
@@ -210,8 +210,7 @@ def __contains__(self, version):
         >>> assert v24 in VersionConstraint(comparator="<=", version=v24)
         >>> assert v24 not in VersionConstraint(comparator="<", version=v24)
         """
-
-        if not isinstance(version, self.version.__class__):
+        if not isinstance(version, self.version.__class__) and self.comparator != "*":
```
pombredanne, line 213: This does not seem to be the root cause.
Instead look at https://github.com/nexB/univers/blob/4dccdfb0041816fb840af9023ff3473490f02829/src/univers/version_range.py#L117
Here we are creating a "star" constraint passing a None class which is bug?

Hritik14: Thanks for pointing out !

PR #34, opened 2022-01-29
Fix for NpmVersionRange.from_native and README

keshav-space:
### 1. Fix `README` 

- Import `versions` instead `version`
     ```python
      [❌] from univers.version import PypiVersion
      [✅] from univers.versions import PypiVersion
   ```
- Use relative-path to execute configure script

### 2. Fix `NpmVersionRange`
- NpmVersionRange.from_native was returning a `tuple` containing `list of VersionConstraint`
- Instead it must return a `tuple` containing `VersionConstraint`

  - Earlier `NpmVersionRange.from_native("^1.0.2")` would return `constraints` as
     ```python
     ([VersionConstraint(comparator='<', version=SemverVersion(string='2.0.0')), VersionConstraint(comparator='>=', version=SemverVersion(string='1.0.2'))],)
    ```
   - After Fix
      ```python
      (VersionConstraint(comparator='>=', version=SemverVersion(string='1.0.2')), VersionConstraint(comparator='<', version=SemverVersion(string='2.0.0')))
     ```
- Fixes: https://github.com/nexB/univers/issues/35

### 3. Add test for `NpmVersionRange`
- `test_NpmVersionRange_from_native_with_compatible_with_version_operator`
- `test_NpmVersionRange_from_native_with_approximately_equal_to_operator`

### 4. Don't do black codestyle test for `docs` and `etc`
- codestyle test for `etc` to be handled by `skeleton`
- see: https://github.com/nexB/skeleton/issues/54

Signed-off-by: keshav-space <keshav6240@gmail.com>

### DIFF
--- a/tests/test_version_range.py
+++ b/tests/test_version_range.py
@@ -13,8 +13,10 @@ from univers.version_range import InvalidVersionRange
 from univers.version_range import PypiVersionRange
 from univers.version_range import VersionRange
 from univers.version_range import RANGE_CLASS_BY_SCHEMES
+from univers.version_range import NpmVersionRange
 from univers.versions import PypiVersion
 from univers.versions import RubygemsVersion
+from univers.versions import SemverVersion
 
 
 class TestVersionRange(TestCase):
@@ -209,9 +211,32 @@ class TestVersionRange(TestCase):
         except InvalidVersionRange as ivre:
             assert str(ivre).startswith("Unsupported character")
 
+    def test_NpmVersionRange_from_native_with_compatible_with_version_operator(self):
+        npm_range = "^1.2.9"
+        expected = NpmVersionRange(
+            constraints=(
+                VersionConstraint(comparator=">=", version=SemverVersion(string="1.2.9")),
+                VersionConstraint(comparator="<", version=SemverVersion(string="2.0.0")),
+            )
+        )
+        version_range = NpmVersionRange.from_native(npm_range)
+        assert version_range == expected
+
+    def test_NpmVersionRange_from_native_with_approximately_equal_to_operator(self):
+        npm_range = "~3.8.2"
+        expected = NpmVersionRange(
+            constraints=(
+                VersionConstraint(comparator=">=", version=SemverVersion(string="3.8.2")),
+                VersionConstraint(comparator="<", version=SemverVersion(string="3.9.0")),
+            )
+        )
+        version_range = NpmVersionRange.from_native(npm_range)
+        assert version_range == expected
+
 
 VERSION_RANGE_TESTS_BY_SCHEME = {
     "nginx": ["0.8.40+", "0.7.52-0.8.39", "0.9.10", "1.5.0+, 1.4.1+"],
+    "npm": ["^1.2.9", "~3.8.2", "5.0.0 - 7.2.3", "2.1 || 2.6", "1.1.2 1.2.2", "<=2.1 >=1.1"],
 }
 
 
--- a/README.rst
+++ b/README.rst
@@ -139,7 +139,7 @@ Compare two native Python versions:
 
 .. code:: python
 
-    from univers.version import PypiVersion
+    from univers.versions import PypiVersion
     assert PypiVersion("1.2.3") < PypiVersion("1.2.4")
 
 
@@ -156,7 +156,7 @@ Test if a version is within or outside a version range:
 
 .. code:: python
 
-    from univers.version import PypiVersion
+    from univers.versions import PypiVersion
     from univers.version_range import VersionRange
 
     range = VersionRange.from_string("vers:pypi/>=1.2.4")
@@ -170,7 +170,7 @@ Development
 
 Run these commands, starting from a git clone of https://github.com/nexB/univers ::
 
-    $ configure --dev
+    $ ./configure --dev
     $ source venv/bin/active
     $ pytest -vvs
 
--- a/src/univers/version_range.py
+++ b/src/univers/version_range.py
@@ -234,10 +234,10 @@ class NpmVersionRange(VersionRange):
         anyof_constraints = []
         if isinstance(clause, AnyOf):
             for allof_clause in clause.clauses:
-                anyof_constraints.append(get_allof_constraints(cls, allof_clause))
+                anyof_constraints.extend(get_allof_constraints(cls, allof_clause))
         elif isinstance(clause, AllOf):
             alloc = get_allof_constraints(cls, clause)
-            anyof_constraints.append(alloc)
+            anyof_constraints.extend(alloc)
         else:
             raise ValueError(f"Unknown clause type: {spec!r}")
 
--- a/tests/test_codestyle.py
+++ b/tests/test_codestyle.py
@@ -10,7 +10,7 @@ import unittest
 
 class BaseTests(unittest.TestCase):
     def test_codestyle(self):
-        args = "venv/bin/black --check -l 100 setup.py etc src tests docs"
+        args = "venv/bin/black --check -l 100 setup.py src tests"
         try:
             subprocess.check_output(args.split())
         except subprocess.CalledProcessError as e:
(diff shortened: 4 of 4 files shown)


pombredanne (changes requested):
Excellent catch!
Thanks... can you add a test or two?

keshav-space:
> Excellent catch! Thanks... can you add a test or two?

Done !!

tests/test_version_range.py
```diff
@@ -209,6 +211,32 @@ def test_PypiVersionRange_raises_ivr_for_invalid_ranges(self):
...
+        # test for `~version` or `Approximately equivalent to version`
+        npm_range = "~3.8.2"
+        npm_range_str = "vers:npm/>=3.8.2|<3.9.0"
+        expected = (
+            VersionConstraint(comparator=">=", version=SemverVersion(string="3.8.2")),
+            VersionConstraint(comparator="<", version=SemverVersion(string="3.9.0")),
+        )
+        version_range = NpmVersionRange.from_native(npm_range)
+        assert version_range.constraints == expected
+        assert version_range.to_string() == npm_range_str
```
Hritik14, line 238: These are two tests clubbed into one. Please split them accordingly. Something like `test_NpmVersionRange_from_native_range_with_caret_operator` `test_NpmVersionRange_from_native_range_with_tilde_operator`
or even better
`test_NpmVersionRange_from_native_range_with_compatible_with_version_operator` and 
`test_NpmVersionRange_from_native_range_with_approximately_equal_to_operator`

tests/test_version_range.py
```diff
@@ -209,6 +211,32 @@ def test_PypiVersionRange_raises_ivr_for_invalid_ranges(self):
         except InvalidVersionRange as ivre:
             assert str(ivre).startswith("Unsupported character")
 
+    def test_NpmVersionRange_from_native(self):
+        """
+        Tests for NpmVersionRange.from_native
+        """
```
Hritik14, line 217: This information is already conveyed by the test name. Maybe a different or no docstring ?

tests/test_version_range.py
```diff
@@ -209,6 +211,32 @@ def test_PypiVersionRange_raises_ivr_for_invalid_ranges(self):
...
+        # test for `^version` or `Compatible with version`
```
Hritik14, line 218: We can say this in the function name itself. Long function names in tests are no problem at all

tests/test_version_range.py
```diff
@@ -209,6 +211,32 @@ def test_PypiVersionRange_raises_ivr_for_invalid_ranges(self):
...
+        npm_range = "^1.2.9"
+        npm_range_str = "vers:npm/>=1.2.9|<2.0.0"
+        expected = (
+            VersionConstraint(comparator=">=", version=SemverVersion(string="1.2.9")),
+            VersionConstraint(comparator="<", version=SemverVersion(string="2.0.0")),
+        )
+        version_range = NpmVersionRange.from_native(npm_range)
+        assert version_range.constraints == expected
+        assert version_range.to_string() == npm_range_str
```
Hritik14, line 227: This test is asserting two different things. How about breaking them into separate tests?

tests/test_version_range.py
```diff
@@ -209,6 +211,32 @@ def test_PypiVersionRange_raises_ivr_for_invalid_ranges(self):
...
+            VersionConstraint(comparator=">=", version=SemverVersion(string="3.8.2")),
+            VersionConstraint(comparator="<", version=SemverVersion(string="3.9.0")),
+        )
+        version_range = NpmVersionRange.from_native(npm_range)
+        assert version_range.constraints == expected
+        assert version_range.to_string() == npm_range_str
+
 
 VERSION_RANGE_TESTS_BY_SCHEME = {
     "nginx": ["0.8.40+", "0.7.52-0.8.39", "0.9.10", "1.5.0+, 1.4.1+"],
```
Hritik14, line 242: As you're already working on npm version range, how about adding an entry here too :)

Hritik14 (changes requested):
Really nice catch! I've added some suggestions for you to consider

tests/test_version_range.py
keshav-space: Wouldn’t  having separate test for `constraints` and `string conversion`  make it cumbersome? 

That would effectively mean for doing small test we would need  these many function all having  almost duplicate codes.

- `test_NpmVersionRange_from_native_range_with_compatible_with_version_operator_constraints`
- `test_NpmVersionRange_from_native_range_with_compatible_with_version_operator_to_string`
- `test_NpmVersionRange_from_native_range_with_approximately_equal_to_operator_constraints`
- `test_NpmVersionRange_from_native_range_with_approximately_equal_to_operator_to_string`

tests/test_version_range.py
keshav-space: Yeah that makes sense.. :)

tests/test_version_range.py
Hritik14: As `NpmVersionRange` does not implement a to_string method, we could avoid testing it now. Simply test it for `VersionRange`. A unit test should ideally be testing a unit only. Testing a group of operations should fall under integration tests

keshav-space: Okay! then going ahead  with `VersionRange` test

tests/test_version_range.py
```diff
@@ -209,9 +211,28 @@ def test_PypiVersionRange_raises_ivr_for_invalid_ranges(self):
...
             assert str(ivre).startswith("Unsupported character")
 
+    def test_NpmVersionRange_from_native_with_compatible_with_version_operator(self):
+        npm_range = "^1.2.9"
+        expected = (
+            VersionConstraint(comparator=">=", version=SemverVersion(string="1.2.9")),
+            VersionConstraint(comparator="<", version=SemverVersion(string="2.0.0")),
+        )
+        version_range = NpmVersionRange.from_native(npm_range)
+        assert version_range.constraints == expected
```
Hritik14, line 221: The function name says about testing `NpmVersionRange_from_native` but you're only testing the constraints.
How about:
- Modifying the function name accordingly or
- Changing the test to comply with the name

tests/test_version_range.py
```diff
@@ -209,9 +211,28 @@ def test_PypiVersionRange_raises_ivr_for_invalid_ranges(self):
...
+
+    def test_NpmVersionRange_from_native_with_approximately_equal_to_operator(self):
+        npm_range = "~3.8.2"
+        expected = (
+            VersionConstraint(comparator=">=", version=SemverVersion(string="3.8.2")),
+            VersionConstraint(comparator="<", version=SemverVersion(string="3.9.0")),
+        )
+        version_range = NpmVersionRange.from_native(npm_range)
+        assert version_range.constraints == expected
```
Hritik14, line 230: Same here

tests/test_version_range.py
```diff
@@ -209,9 +211,28 @@ def test_PypiVersionRange_raises_ivr_for_invalid_ranges(self):
...
+
 
 VERSION_RANGE_TESTS_BY_SCHEME = {
     "nginx": ["0.8.40+", "0.7.52-0.8.39", "0.9.10", "1.5.0+, 1.4.1+"],
+    "npm": ["^1.2.9", "~3.8.2"],
```
Hritik14, line 235: Can you please add more ranges for npm ? It would really help make other tests better.

Hritik14 (changes requested):
Looks good! Just a few changes to go

tests/test_version_range.py
keshav-space: Yeah, but that's what `from_native` is doing.. i.e returning  tuple of `VersionConstraint`.
And the  test `NpmVersionRange_from_native` only intendeds to test that specific `constraints` tuple.
Right?

keshav-space: Yeah, That makes complete sense to me.
Will modify the tests to do assertion for `NpmVersionRange` instead of `constraints`

And I guess we have a similar issue with `GemVersionRange_from_native`, shall we do something about this too?

https://github.com/nexB/univers/blob/21864169341c4949bd40c931cd8450488de24d64/tests/test_version_range.py#L182-L189

Hritik14: With tests, it's always the more the merrier :)

Hritik14:
@keshav-space LGTM! 
Please squash your commits in two parts - black formatting and the fix.

pombredanne (approved):
LGTM. Thanks!
Merging!

Issue #36, opened 2022-02-07
Create new version range for "openssl"


pombredanne:
See https://www.openssl.org/policies/releasestrat.html

There are still old 1.1.1-style versions and these look like a lot PyPI.

keshav-space:
For the base version `3.0` onward OpenSSL is using Semver
and for `1.1.1` they will continue with old versioning style ( 1.1.1 will be supported until 2023-09-11)

> Rest all other versions are no longer publicly supported


So, for openssl we will have two different versioning schema depending on the version ;(
>Imagine Something like this `vers:openssl/1.1.1k|3.0.2`

Only thing that seems good ->  https://www.openssl.org/news/vulnerabilities.xml

pombredanne:
> So, for openssl we will have two different versioning schema depending on the version ;(

Yes! what I suggest as a design is this:
Something like a single OpensslVersion class that behaves differently if this is a 3 or an earlier version and once constructed track an internal version object that it delegates to:: either an old style version (string or some new LegacyOpensslVersion object)  or a Semver version object

Issue #41, opened 2022-02-21
Treat `OpenSSL` and `OpenSSL-FIPS Module` as a seperate initity


keshav-space:
`OpenSSL-FIPS Object Module` is going nowhere it's here to stay
see: https://wiki.openssl.org/index.php/FIPS_module_3.0
https://www.youtube.com/watch?v=7_GVQNJA15I
Currently available fips:
- fips-1.0
- fips-1.1
- fips-1.2
- fips-2.0

https://www.openssl.org/docs/manmaster/man7/fips_module.html

PR #42, opened 2022-02-21
Add openssl support in univers

keshav-space:
- closes https://github.com/nexB/univers/issues/36
- For `OpenSSL-FIPS Module` see https://github.com/nexB/univers/issues/41

Signed-off-by: Keshav Priyadarshi <git@keshav.space>


src/univers/versions.py
```diff
@@ -343,3 +344,86 @@ def __gt__(self, other):
         if not isinstance(other, self.__class__):
             return NotImplemented
         return gentoo.vercmp(self.value, other.value) > 0
+
+
+@attr.s(frozen=True, order=False, eq=False, hash=True)
+class LegacyOpensslVersion(Version):
```
Hritik14, line 350:
```suggestion
class OpensslLegacyVersion(Version):
```

src/univers/versions.py
```diff
@@ -343,3 +344,86 @@ def __gt__(self, other):
...
+    def __ge__(self, other):
+        return self.__gt__(other) or self.__eq__(other)
+
+
+@attr.s(frozen=True, order=False, eq=False, hash=True)
+class OpensslVersion:
+    """
+    Handles the object creation for openssl depending on
+    whether version is legacy or semver.
+    """
```
Hritik14, line 402:
```suggestion
    """
    Openssl versions are of two types
        - Legacy versions: Implemented in OpensslLegaryVersion
        - New versions: Semver
    """
```

Hritik14: Also, have they explicitly named these version types as ``Legacy`` and ``New`` ?

src/univers/versions.py
```diff
@@ -343,3 +344,86 @@ def __gt__(self, other):
...
+
+    def __new__(cls, string):
```
Hritik14, line 404:
```suggestion
    def __new__(cls, string):
    """
    Handle the object creation for Openssl version depending on
    whether it is legacy or new/semver.
    """
```

src/univers/versions.py
```diff
@@ -343,3 +344,86 @@ def __gt__(self, other):
...
+            return LegacyOpensslVersion(string)
+        elif OpensslVersion.is_valid_new_openssl(string):
+            return SemverVersion(string)
+        raise InvalidVersion(f"{string!r} is not a valid/supported OpensslVersion")
+
+    @classmethod
+    def is_valid_legacy(cls, string):
+        """
+        All legacy base version of openssl that ever exited/exists.
+        """
```
Hritik14, line 415: This should be a comment

src/univers/versions.py
```diff
@@ -343,3 +344,86 @@ def __gt__(self, other):
...
+        all_legacy_base = ["0.9.6", "0.9.7", "0.9.8", "1.0.0", "1.0.1", "1.0.2", "1.1.0", "1.1.1"]
+        if string in all_legacy_base:
+            return True
+        regexed_string = re.findall(r"^(\d+\.\d+\.\d+)([a-z-]+)$", string)
```
Hritik14, line 419: This is hard to read, can we have examples ? Examples in the form of doctests would be perfect!

src/univers/versions.py
```diff
@@ -343,3 +344,86 @@ def __gt__(self, other):
...
+        if not regexed_string or regexed_string[0][0] not in all_legacy_base:
+            return False
+        return True
+
+    @classmethod
+    def is_valid_new_openssl(cls, string):
+        regexed_string = re.findall(r"^(\d+)(\.)(.*)$", string)
```
Hritik14, line 426: Example and doctests here too please

src/univers/versions.py
```diff
@@ -343,3 +344,86 @@ def __gt__(self, other):
...
+
+@attr.s(frozen=True, order=False, eq=False, hash=True)
+class LegacyOpensslVersion(Version):
+    @classmethod
+    def is_valid(cls, string):
+        return OpensslVersion.is_valid_legacy(string)
+
+    def __eq__(self, other):
+        if isinstance(other, SemverVersion):
+            return False
```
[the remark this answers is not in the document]
Hritik14: I see it now. Can you add a few examples of legacy openssl versions in the docstring ?

src/univers/versions.py
```diff
@@ -343,3 +344,86 @@ def __gt__(self, other):
         if not isinstance(other, self.__class__):
             return NotImplemented
         return gentoo.vercmp(self.value, other.value) > 0
+
+
+@attr.s(frozen=True, order=False, eq=False, hash=True)
+class LegacyOpensslVersion(Version):
+    @classmethod
+    def is_valid(cls, string):
+        return OpensslVersion.is_valid_legacy(string)
```
Hritik14, line 353: Implement this here and call it from OpensslVersion (ie the other way round)

src/univers/versions.py
```diff
@@ -343,3 +344,86 @@ def __gt__(self, other):
...
+        if isinstance(other, SemverVersion):
+            return False
+        if not isinstance(other, self.__class__):
+            return NotImplemented
+        return self.value == other.value
+
+    def __lt__(self, other):
+        """
+        All legacy scheme is behind those using semver.
+        """
```
Hritik14, line 365:
```suggestion
```

src/univers/versions.py
```diff
@@ -343,3 +344,86 @@ def __gt__(self, other):
...
+        if isinstance(other, SemverVersion):
```
Hritik14, line 366:
```suggestion
        if isinstance(other, SemverVersion):
            # SemverVersion is the new openssl scheme, thus always greater than legacy
```

src/univers/versions.py
```diff
@@ -343,3 +344,86 @@ def __gt__(self, other):
...
+        re_vers1 = re.findall(r"^(\d+\.\d+\.\d+)(.*)$", self.value)[0]
+        re_vers2 = re.findall(r"^(\d+\.\d+\.\d+)(.*)$", other.value)[0]
+        if int(re_vers1[0].replace(".", "")) < int(re_vers2[0].replace(".", "")):
+            return True
+        return re_vers1[0] == re_vers2[0] and re_vers1[1] < re_vers2[1]
+
+    def __gt__(self, other):
+        """
+        All openssl version using semver is ahead of legacy scheme.
+        """
```
Hritik14, line 379:
```suggestion
```

src/univers/versions.py
```diff
@@ -343,3 +344,86 @@ def __gt__(self, other):
...
+        if isinstance(other, SemverVersion):
```
Hritik14, line 380:
```suggestion
        if isinstance(other, SemverVersion):
            # SemverVersion is the new openssl scheme, thus always greater than legacy
```

src/univers/versions.py
```diff
@@ -343,3 +344,86 @@ def __gt__(self, other):
...
+            return True
+        if not isinstance(other, self.__class__):
+            return NotImplemented
+        re_vers1 = re.findall(r"^(\d+\.\d+\.\d+)(.*)$", self.value)[0]
+        re_vers2 = re.findall(r"^(\d+\.\d+\.\d+)(.*)$", other.value)[0]
+        if int(re_vers1[0].replace(".", "")) < int(re_vers2[0].replace(".", "")):
+            return True
+        return re_vers1[0] == re_vers2[0] and re_vers1[1] < re_vers2[1]
```
Hritik14, line 374: This is cryptic. Can you make it simpler ?

src/univers/versions.py
```diff
@@ -343,3 +344,86 @@ def __gt__(self, other):
...
+        """
+        if isinstance(other, SemverVersion):
+            return False
+        if not isinstance(other, self.__class__):
+            return NotImplemented
+        re_vers1 = re.findall(r"^(\d+\.\d+\.\d+)([a-z-]+)$", self.value)[0]
+        re_vers2 = re.findall(r"^(\d+\.\d+\.\d+)([a-z-]+)$", other.value)[0]
+        if int(re_vers1[0].replace(".", "")) > int(re_vers2[0].replace(".", "")):
+            return True
+        return re_vers1[0] == re_vers2[0] and re_vers1[1] > re_vers2[1]
```
Hritik14, line 388: Hard to understand => prone to errors

Hritik14 (changes requested):
Thanks for the research on openssl!! There are a few things that I'd like you to consider...

src/univers/version_constraint.py
```diff
@@ -157,6 +157,22 @@ def from_string(cls, string, version_class):
...
+        Example:
+        OpensslVersion when instantiated will return either an object of
+        LegacyOpensslVersion or SemverVersion depending on version.
+        OpensslVersion("0.9.8a") -> LegacyOpensslVersion(string="0.9.8a")
+        OpensslVersion("3.0.1") -> SemverVersion(string="3.0.1")
+
+        Hence the need for updating `version_class` after creation of version object.
+        """
+        if version != None:
+            version_class = version.__class__
```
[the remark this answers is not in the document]
keshav-space: It's not about having an empty version_class, but about having the wrong version_class.
In events when version_class returns a different class object we need to update the version_class accordingly. But of course, if the version is `None` (as in the case when comparator == "*"  ) in that case no Version Object is instantiated hence no need to update the `version_class`

let's say we have something like this:
```python
VersionConstraint.from_string("1.0.1f", OpensslVersion)
```
https://github.com/nexB/univers/blob/7206c1fc35eb32d0006591e6defcc65ded40cfa0/src/univers/version_constraint.py#L159-L160
On execution on L159 version will be like: 
(Note: OpensslVersion would return either `LegacyOpensslVersion` or `SemverVersion` depending on version string)
```python
LegacyOpensslVersion(string="1.0.1f")
```
Return on L160 would be like:
```python
return cls(comparator="=", version=LegacyOpensslVersion(string="1.0.1f"), version_class=OpensslVersion) 
```
So, we end up with an absurd return statement where version -> `LegacyOpensslVersion` but version_class is still `OpensslVersion` and that will go on to create problem

src/univers/versions.py
keshav-space: So in the case of naming, I borrowed it from Philips's suggestion.
https://github.com/nexB/univers/issues/36#issuecomment-1034629474

src/univers/versions.py
keshav-space: No, they don't call it new or Legacy.. they're just calling it old and contemporary format..🤷

src/univers/versions.py
keshav-space: Yeah that makes more sense 😆

src/univers/versions.py
keshav-space: Is this good enough??
```python
regexed_vers_self = re.findall(r"^(\d+\.\d+\.\d+)(.*)$", self.value)[0]
regexed_vers_other = re.findall(r"^(\d+\.\d+\.\d+)(.*)$", other.value)[0]

base_vers_self = regexed_version_self[0]
base_vers_other = regexed_version_other[0]

numeric_base_version_self = int(base_vers_self.replace(".","")
numeric_base_version_other = int(base_vers_other.replace(".","")

postfix_self = regexed_version_self[1]
postfix_other = regexed_version_other[1]

if numeric_base_version_self < numeric_base_version_other:
      return True
return base_vers_self == base_vers_other and postfix_self < postfix_other

```

src/univers/version_constraint.py
```diff
@@ -157,6 +157,22 @@ def from_string(cls, string, version_class):
...
+        of its own kind it may return an object of a different class.
+
+        Example:
+        OpensslVersion when instantiated will return either an object of
+        LegacyOpensslVersion or SemverVersion depending on version.
+        OpensslVersion("0.9.8a") -> LegacyOpensslVersion(string="0.9.8a")
+        OpensslVersion("3.0.1") -> SemverVersion(string="3.0.1")
+
+        Hence the need for updating `version_class` after creation of version object.
+        """
```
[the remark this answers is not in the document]
keshav-space: Shall I remove this altogether?

src/univers/version_range.py
```diff
@@ -868,6 +868,28 @@ def from_native(cls, string):
         return cls(constraints=constraints)
 
 
+class OpensslVersionRange(VersionRange):
+    """
+    openssl doesn't use <,>,<= or >=.
+    For more 'https://www.openssl.org/news/vulnerabilities.xml'.
```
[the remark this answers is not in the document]
keshav-space: will document that..👍🏼

src/univers/version_range.py
```diff
@@ -868,6 +868,39 @@ def from_native(cls, string):
...
+    ... VersionConstraint(comparator="=", version=OpensslVersion("3.0.1")),
+    ... VersionConstraint(comparator="=", version=OpensslVersion("1.1.1nf")),
+    ... )
+    >>> range = OpensslVersionRange(constraints=constraints)
+    >>> assert str(range) == 'vers:openssl/1.0.1af|1.1.1nf|3.0.1'
+    """
+
+    scheme = "openssl"
+    version_class = versions.OpensslVersion
+    vers_by_native_comparators = {"=": "="}
```
pombredanne, line 890: Do you need this?

src/univers/version_range.py
```diff
@@ -868,6 +868,39 @@ def from_native(cls, string):
...
+
+    @classmethod
+    def from_native(cls, string):
+        cleaned = remove_spaces(string).lower()
+        constraints = set()
```
pombredanne, line 895: Why a set and not a list?

src/univers/versions.py
```diff
@@ -343,3 +343,173 @@ def __gt__(self, other):
...
+
+
+@attr.s(frozen=True, order=False, eq=False, hash=True)
+class LegacyOpensslVersion(Version):
+    """
+    Represent an Legacy Openssl Version .
+
+    For example::
+
+    # 1.0.1f|0.9.7d|1.0.2ac
```
pombredanne, line 355: Make this a docstring instead

src/univers/versions.py
```diff
@@ -343,3 +343,173 @@ def __gt__(self, other):
...
+    """
+
+    @classmethod
+    def is_valid(cls, string):
+        return bool(cls.parse(string))
+
+    @classmethod
+    def parse(cls, string):
+
```
pombredanne, line 364:
```suggestion
```

src/univers/versions.py
pombredanne, line 355: Make this a doctest instead

src/univers/versions.py
```diff
@@ -343,3 +343,173 @@ def __gt__(self, other):
...
         return gentoo.vercmp(self.value, other.value) > 0
+
+
+@attr.s(frozen=True, order=False, eq=False, hash=True)
+class LegacyOpensslVersion(Version):
+    """
+    Represent an Legacy Openssl Version .
+
+    For example::
+
```
pombredanne, line 354:
```suggestion
```

src/univers/versions.py
```diff
@@ -343,3 +343,173 @@ def __gt__(self, other):
...
+    @classmethod
+    def is_valid(cls, string):
+        return bool(cls.parse(string))
+
+    @classmethod
+    def parse(cls, string):
+
+        """
+        Returns the tuple containig the 4 segments (i.e major, minor, build, patch) of Legacy Version,
```
pombredanne, line 366:
```suggestion
        Return a  four-tuple of (major, minor, build, patch) version segments where
        major, minor, build are integers and patch is a trsing possibly empty.
        Return False if this is not a valid LegacyOpensslVersion.
```

src/univers/versions.py
```diff
@@ -343,3 +343,173 @@ def __gt__(self, other):
...
+        For example::
+        >>> LegacyOpensslVersion.parse("1.0.1f")
+        (1, 0, 1, 'f')
+        >>> LegacyOpensslVersion.parse("1.0.2ac")
+        (1, 0, 2, 'ac')
+        >>> LegacyOpensslVersion.parse("2.0.2az")
+        False
+        """
+
+        # All legacy base version of openssl that ever exited/exists.
```
pombredanne, line 378:
```suggestion
        # All known legacy base OpenSLL versions
```

src/univers/versions.py
```diff
@@ -343,3 +343,173 @@ def __gt__(self, other):
...
+            "0.9.6",
+            "0.9.7",
+            "0.9.8",
+            "1.0.0",
+            "1.0.1",
+            "1.0.2",
+            "1.1.0",
+            "1.1.1",
+        )
+        # check if starts with a valid base
```
pombredanne, line 394: This comment is not needed...
`check if starts with a valid base` does not explain anything extra than the code `if not string.startswith(all_legacy_base)` does.
When the code is readable and obvious, comments are not needed.

src/univers/versions.py
```diff
@@ -343,3 +343,173 @@ def __gt__(self, other):
...
+        major, minor, build = segments
+        major = int(major)
+        minor = int(minor)
+        if build.isdigit():
+            build = int(build)
+            patch = ""
+        else:
+            patch = build[1:]
+            build = int(build[0])
+        if patch and patch[0].isdigit():
```
pombredanne, line 410:
```suggestion
             if patch and patch[0].isdigit():
```
this should be indented as this is never needed when `patch=""`

src/univers/versions.py
```diff
@@ -343,3 +343,173 @@ def __gt__(self, other):
...
+            return False
+        return major, minor, build, patch
+
+    @classmethod
+    def build_value(cls, string):
+        return cls.parse(string)
+
+    def __str__(self):
+        return self.normalized_string
```
pombredanne, line 419: What about instead:
```suggestion
        return f"{self.major}.{self.minor}.{self.build}{self.patch}"
```

src/univers/versions.py
```diff
@@ -343,3 +343,173 @@ def __gt__(self, other):
...
+
+
+@attr.s(frozen=True, order=False, eq=False, hash=True)
+class OpensslVersion(Version):
+
```
pombredanne, line 424:
```suggestion
```

src/univers/versions.py
```diff
@@ -343,3 +343,173 @@ def __gt__(self, other):
...
+    """
+    Internally tracks two types of openssl versions
+        - Legacy versions: Implemented in LegacyOpensslVersion
```
pombredanne, line 427:
```suggestion
        - LegacyOpensslVersion for versions before version 3.0.0 such as 1.0.1g
```

src/univers/versions.py
```diff
@@ -343,3 +343,173 @@ def __gt__(self, other):
...
+        - New versions: Semver
```
pombredanne, line 428:
```suggestion
        - Semver for versions from 3.0.0 and up
```

src/univers/versions.py
```diff
@@ -343,3 +343,173 @@ def __gt__(self, other):
...
+        """
+        if cls.is_valid_legacy(string):
+            return LegacyOpensslVersion(string)
+        if cls.is_valid_new(string):
+            return SemverVersion(string)
+
+    @classmethod
+    def is_valid_new(cls, string):
+        """
+        Checks the validity of new Openssl Version.
```
pombredanne, line 460: Always use imperative style.
```suggestion
        Check the validity of new Openssl Version.
```

src/univers/versions.py
```diff
@@ -343,3 +343,173 @@ def __gt__(self, other):
...
+        """
+        if SemverVersion.is_valid(string):
+            sem = semantic_version.Version.coerce(string)
+            return sem.major >= 3
+
+    @classmethod
+    def is_valid_legacy(cls, string):
+        return LegacyOpensslVersion.is_valid(string)
+
+    def __eq__(self, other):
```
pombredanne, line 478: You should add tests for all the OpensslVersion version comparisons and these should be testing a matrix of:
- all of >,<, ==, !=, >= and <=
- with new-new, old-new, new-old, old-old  and non OpensslVersion object values, as well as LegacyOpensslVersion and Semver versions 
- in the general case and right at the boundaries of legacy and new

pombredanne (changes requested):
Thank you ++
See some comments inline for your consideration.

src/univers/version_range.py
keshav-space: No, not really needed.

src/univers/version_range.py
keshav-space: Just to avoid having possible duplicates

src/univers/versions.py
keshav-space: Indeed, will replace with
```python
        else:
            patch = build[1:]
            build = int(build[0])
            if patch[0].isdigit():
```

src/univers/versions.py
keshav-space: Will add that

src/univers/version_range.py
pombredanne: remove it then:

```suggestion
```

src/univers/version_range.py
```diff
@@ -868,6 +868,39 @@ def from_native(cls, string):
         return cls(constraints=constraints)
 
 
+class OpensslVersionRange(VersionRange):
+    """
+    Openssl version range.
+    openssl doesn't use <,>,<= or >=
+    For more see 'https://www.openssl.org/news/vulnerabilities.xml'
+
+    For exmaple::
```
pombredanne, line 877:
```suggestion
    For example::
```

src/univers/version_range.py
pombredanne: why would there be possible duplicates? There are no other VersionRange that uses sets this way for constraints.
It is likely a good idea to avoid duplicated constraints (and I should add this to the vers spec)  but it should not be the responsability of a subclass... rather it would be done in one single place here https://github.com/nexB/univers/blob/6f03cf67d094fd823aca72bbb5352ccfcb0936fe/src/univers/version_range.py#L53 and properly tested. Let's do this in another PR... can you enter an issue for this?

src/univers/version_range.py
```diff
@@ -868,6 +868,39 @@ def from_native(cls, string):
...
+
+    scheme = "openssl"
+    version_class = versions.OpensslVersion
+    vers_by_native_comparators = {"=": "="}
+
+    @classmethod
+    def from_native(cls, string):
+        cleaned = remove_spaces(string).lower()
+        constraints = set()
+        # plain single version
```
pombredanne, line 896: What does this comment wants to tell me?

src/univers/version_range.py
```diff
@@ -868,6 +868,39 @@ def from_native(cls, string):
...
+        for clause in cleaned.split(","):
```
pombredanne, line 897: why call this a clause if this is a version?

src/univers/versions.py
pombredanne, line 354: You do not need an empty line there... if you mark this as resolved, then remove the line ;)
```suggestion
```

src/univers/versions.py
pombredanne: Same ... this is marked resolved but this has no changed?

pombredanne (changes requested):
Hey :) I added a few more comments... also I did unresolve a few of the comments you had marked as resolved as I could not see the corresponding change applied... I realize now thst this may be just because you have not pushed your changes yet?

src/univers/version_range.py
keshav-space: Yeah, I have implemented all those suggestions locally.. once the test for version comparison is ready I will push those changes

src/univers/version_range.py
keshav-space: Just make it look consistent across the version range implementations... Yeah but I get that it's just a version

src/univers/versions.py
keshav-space: Seems more pragmatic...👍

src/univers/version_range.py
keshav-space: Frankly, it's just stating the obvious

keshav-space: Frankly, it's just stating the obvious.

src/univers/version_range.py
keshav-space: Thanks, then i don't see any chance for possible duplication. Will create an issue for adding some tests there.

pombredanne (commented):
If you merge the latest main back in your branch the tests should not run anymore on older mac/linux/windows (and should therefore no longer fail)

src/univers/versions.py
keshav-space: Won't work.
```python
>>> from univers.versions import LegacyOpensslVersion
>>> vers=LegacyOpensslVersion("1.0.2a")
>>> str(vers)
Traceback (most recent call last):
    ...
AttributeError: 'LegacyOpensslVersion' object has no attribute 'major'
```

We can have something like this
```python
return f"{self.value[0]}.{self.value[1]}.{self.value[2]}{self.value[3]}"
```

pombredanne (approved):
Looking great. We can merge as is...Thank you ++
Can you please add an issue to add an extra test that would use the list of all the OpenSSL version ever released to date and ensure that sorting this list gives the correct order?

keshav-space:
> Looking great. We can merge as is...Thank you ++ Can you please add an issue to add an extra test that would use the list of all the OpenSSL version ever released to date and ensure that sorting this list gives the correct order?

Yeah, I will raise an issue for that.

PR #55, opened 2022-03-27
add `from_sequence` in OpensslVersionRange

keshav-space:
context: https://github.com/nexB/vulnerablecode/pull/648#discussion_r834011600
Signed-off-by: Keshav Priyadarshi <git@keshav.space>


TG1999:
Thanks @keshav-space , this looks good to me, @pombredanne should we shift this to the VersionRange class so other sub classes can also use this functionality, to form ranges from sequences ?

src/univers/version_range.py
```diff
@@ -898,6 +898,16 @@ def from_native(cls, string):
             constraints.append(constraint)
         return cls(constraints=constraints)
 
+    @classmethod
+    def from_sequence(cls, sequence):
```
pombredanne, line 902: Could you add a docstring?
what is this doing? what is the input and what is this returning?
May be `Return a version range from a list of version strings.`

Also is `from_sequence` the best name? what is the use case here? I would say this is to create a range from a list of versions? what about `from_versions`

src/univers/version_range.py
```diff
@@ -898,6 +898,16 @@ def from_native(cls, string):
...
+        constraints = []
+        for version in sequence:
+            version = remove_spaces(version).lower()
```
TG1999, line 905: VersionClass constructor should take care of this

Issue #47, opened 2022-03-14
Add support for from_native function


TG1999:
Add support for `from_native` function for following version ranges:
- [ ] Maven
- [ ] Nuget
- [ ] Composer
- [ ] GoLang

pombredanne:
In doing so, we want to also use that in PurlDB, ScanCode and VulnerableCode.

PR #58, opened 2022-04-01
Add version range support for maven #47

TG1999:

### DIFF
--- a/src/univers/version_range.py
+++ b/src/univers/version_range.py
@@ -12,6 +12,7 @@ from semantic_version.base import AllOf
 from semantic_version.base import AnyOf
 
 from univers import gem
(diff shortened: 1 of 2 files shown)


pombredanne (changes requested):
Great! just a few nits
We should also test with multiple ranges  like in `affected_range: "[2.0,2.3.1],[2.4.0,2.12.2),[2.13.0,2.15.0)"` from https://gitlab.com/gitlab-org/security-products/gemnasium-db/-/blob/master/maven/org.apache.logging.log4j/log4j-core/CVE-2021-44228.yml#L20

pombredanne (approved):
Looking great! :+1:

PR #60, opened 2022-04-12
Add Nuget Version support in Univers

TG1999:


TG1999:
TODO:
Add ABOUT file for Nuget code
Add NugetVersion Range and add tests for it

src/univers/nuget.py.ABOUT
```diff
@@ -0,0 +1,11 @@
+about_resource: maven.py
```
pombredanne, line 1: I am not sure you meant maven?

src/univers/nuget.py.ABOUT
```diff
@@ -0,0 +1,11 @@
...
+package_url: pkg:github/google/osv
+copyright: |
+    Copyright 2022 Google LLC
+
+notes: This has been substantially modified and enhanced from the original 
+ pymaven code to extract the version comparison code.
+
```
pombredanne, line 8: We need a direct download URL including a commit hash

src/univers/nuget.py.NOTICE
```diff
@@ -0,0 +1,13 @@
+# Copyright 2022 Google LLC
```
pombredanne, line 1: We likely want this also in the nuget.py.

src/univers/nuget.py
```diff
@@ -0,0 +1,163 @@
+#
+# Copyright (c) nexB Inc. and others.
```
pombredanne, line 2: Is this rather copyright Google? You probably want an initial commit with the original unmodified files? And did you modify this code?
I see that https://github.com/google/osv/blob/f5647ad2f746685b08debfba0293e442f2fb9945/lib/osv/semver_index.py#L1 and https://github.com/google/osv/blob/f5647ad2f746685b08debfba0293e442f2fb9945/lib/osv/nuget.py both have a notice and we surely do not want to remove their copyrights, even if this is still in the ABOUT and NOTICE file

src/univers/nuget.py
```diff
@@ -0,0 +1,163 @@
...
+# SPDX-License-Identifier: Apache-2.0
+#
+# Visit https://aboutcode.org and https://github.com/nexB/univers for support and download.
+
+import functools
+import re
+
+import semver
```
pombredanne, line 10: Could we eventually avoid having two semver libraries and reuse `semantic-version` we already use here?
 @rbarrois https://pypi.org/project/semantic-version/ https://github.com/rbarrois/python-semanticversion that we use can parse node-semver version specs constraints while https://github.com/python-semver/python-semver/ can only deal with versions and not the npm specs (we use to convert from npm native to our vers syntax)  which is why we picked in the first place.

pombredanne (changes requested):
Thanks! ... here are some nits for your consideration!

src/univers/nuget.py.ABOUT
TG1999: Oops, fixing this

src/univers/nuget.py
TG1999: I tried reusing the existing library at first place, It returns prerelease and build as tuples, whereas the newly imported one returns a string, and there is no concrete way to get to strings from those tuples.
For example:
```
semantic_version.Version.coerce("1.0.0.0").prerelease
('0',)
nuget.Version.from_string("1.0.0.0").prerelease
None
```
That's why I think we should use the new library for nuget

TG1999: One more thing I observed here, the equal comparison of both the libraries are different, I tried to run all the tests after making the required adjustments by using the `semantic_version` library, but some corner cases are failing there, one more reason to use the new library for nuget

TG1999: Here is the gist for same https://gist.github.com/TG1999/724c868fa21882aa880c091cf5e1964d

TG1999: "1.0.0", "1.0.0+beta"

pombredanne: > "1.0.0", "1.0.0+beta"

So if I get you, nuget consider that these tow are equal? https://github.com/nexB/univers/pull/60/files#diff-1b35636a23b0127a5ed3b726fe14efdf9df1a159e27dd87ead2be71547b78cccR39

This is mind blowing!

*sigh* ok, well, so let's use this for now. :|

tests/test_nuget.py
```diff
@@ -0,0 +1,77 @@
+# Copyright 2022 Google LLC
+#
```
pombredanne, line 2: Adding the missing copyright from upstream https://github.com/NuGet/NuGet.Client/blob/eea4636696cc7b6a3e682797c3dd717ad25fd7c9/test/NuGet.Core.Tests/NuGet.Versioning.Test/VersionComparerTests.cs
```suggestion
# Copyright (c) .NET Foundation. All rights reserved.
```

pombredanne (approved):
LGTM! Thanks!

Issue #46, opened 2022-03-14
test the sorting of all the OpenSSL version ever released


keshav-space:
add a new test to verify the sorting of all the OpenSSL versions ever released.
context: https://github.com/nexB/univers/pull/42#pullrequestreview-908108533

PR #61, opened 2022-04-17
Test sorting of all the OpenSSL versions ever released

keshav-space:
- fix OpenSSL pre-release comparison
- test the sorting of all the OpenSSL versions ever released
- closes: https://github.com/nexB/univers/issues/46


src/univers/versions.py
```diff
@@ -447,6 +446,33 @@ def build_value(cls, string):
...
+        base_vers2 = other.value[:-1]
+        if base_vers1 == base_vers2 and (
+            self.is_prerelease(self.value) != self.is_prerelease(other.value)
+        ):
+            return self.is_prerelease(other.value)
+        return self.value.__gt__(other.value)
+
+    def is_prerelease(self, value):
+        patch = value[-1]
+        return patch.startswith("-beta") or patch.startswith("-alpha")
```
pombredanne, line 474: This is cleaner and means the same:

```suggestion
        return patch.startswith(("-beta","-alpha",))
```

src/univers/versions.py
```diff
@@ -447,6 +446,33 @@ def build_value(cls, string):
...
+            return NotImplemented
+        base_vers1 = self.value[:-1]
+        base_vers2 = other.value[:-1]
+        if base_vers1 == base_vers2 and (
+            self.is_prerelease(self.value) != self.is_prerelease(other.value)
+        ):
+            return self.is_prerelease(other.value)
+        return self.value.__gt__(other.value)
+
+    def is_prerelease(self, value):
```
pombredanne, line 472: What passing a value there? isn't this something to get directtly from self instead, especially since "patch" would become an attribute of self, rather than the [-1] positional index in some tuple

src/univers/versions.py
```diff
@@ -447,6 +446,33 @@ def build_value(cls, string):
     def __str__(self):
         return f"{self.value[0]}.{self.value[1]}.{self.value[2]}{self.value[3]}"
 
+    def __lt__(self, other):
+        if not isinstance(other, self.__class__):
+            return NotImplemented
+        base_vers1 = self.value[:-1]
```
pombredanne, line 452: Rather than having a positional tuple, why not use instead a proper object for this? an attr class with major, minor, build, and patch attributes would go a long way and would avoid having to deal with the magic [:-1]?
Or more simply, just define major, minor, build, and patch as attributes on this object itself. This would be much more explicit and cleaner IMHO.... Always prefer names over positional indexes.

pombredanne (changes requested):
Thanks ... See a few nits for you consideration

src/univers/versions.py
```diff
@@ -385,6 +384,19 @@ class LegacyOpensslVersion(Version):
...
+    patch = attr.ib(type=str, default=None, repr=False)
+
+    def __attrs_post_init__(self):
+        Version.__attrs_post_init__(self)
+        major, minor, build, patch = self.parse(self.string)
+        object.__setattr__(self, "major", major)
+        object.__setattr__(self, "minor", minor)
+        object.__setattr__(self, "build", build)
+        object.__setattr__(self, "patch", patch)
+
```
pombredanne, line 399: I am not sure I like double parsing the string and overriding __attrs_post_init__ without calling super() too?
What about this code instead? (it is not 100% satisfying, but at least it does not double parse):
```
	
    @property
    def major(self):
        major, _minor, _build, _patch = self.value
        return major

    @property
    def minor(self):
        _major, minor, _build, _patch = self.value
        return minor

    @property
    def build(self):
        _major, _minor, build, _patch = self.value
        return build

    @property
    def patch(self):
        _major, _minor, _build, patch = self.value
        return patch
```
or at least reusing the self.value rather than reparsing it?

pombredanne (changes requested):
LGTM... with a minor nitpicking

src/univers/versions.py
pombredanne: LGTM :+1:

pombredanne (approved):
LGTM! Merging!

PR #63, opened 2022-04-26
Subclass semver for nginx

pombredanne:
Because nginx versions are not exactly semver.

Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>

### DIFF
--- a/src/univers/versions.py
+++ b/src/univers/versions.py
@@ -160,6 +160,7 @@ class Version:
 
 @attr.s(frozen=True, order=False, hash=True)
 class GenericVersion(Version):
+
     @classmethod
     def is_valid(cls, string):
         # generic implementation ...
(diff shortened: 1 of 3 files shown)


src/univers/versions.py
```diff
@@ -219,6 +219,64 @@ def is_valid(cls, string):
         except ValueError:
             return False
 
+    @property
+    def major(self):
+        return self.value and self.value.major
```
Hritik14, line 224: Nice trick !

TG1999 (approved):
LGTM!

src/univers/versions.py
pombredanne: That's more of an idiom to me! Though in earnest I am always split in abusing this shortcutting of "and" and "or" and their returning of a value as used here ... it can become hard to read unless used for very simple checks like this one.

PR #64, opened 2022-04-27
Fix nuget string representation

TG1999:

### DIFF
--- a/tests/test_version_range.py
+++ b/tests/test_version_range.py
@@ -9,6 +9,7 @@ import pytest
 
 from univers.version_constraint import VersionConstraint
 from univers.version_range import GemVersionRange
+from univers.version_range import NugetVersionRange
 from univers.version_range import InvalidVersionRange
 from univers.version_range import PypiVersionRange
 from univers.version_range import VersionRange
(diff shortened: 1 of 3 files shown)


tests/test_version_range.py
```diff
@@ -8,14 +8,14 @@
 import pytest
 
 from univers.version_constraint import VersionConstraint
-from univers.version_range import GemVersionRange
+from univers.version_range import GemVersionRange, NugetVersionRange
```
pombredanne, line 11:
```suggestion
from univers.version_range import GemVersionRange
from univers.version_range import NugetVersionRange
```

tests/test_version_range.py
```diff
@@ -8,14 +8,14 @@
...
 from univers.version_range import InvalidVersionRange
 from univers.version_range import PypiVersionRange
 from univers.version_range import VersionRange
 from univers.version_range import RANGE_CLASS_BY_SCHEMES
 from univers.version_range import NpmVersionRange
 from univers.version_range import OpensslVersionRange
-from univers.versions import PypiVersion
+from univers.versions import NugetVersion, PypiVersion
```
pombredanne, line 18:
```suggestion
from univers.versions import NugetVersion
from univers.versions import PypiVersion
```

pombredanne (approved):
LGTM... with just a tiny nit as I prefer to keep one import per line.
Please apply and merge!

PR #68, opened 2022-05-10
Add support for gitlab in pypi, npm, gem version ranges

TG1999:

### DIFF
--- a/src/univers/version_range.py
+++ b/src/univers/version_range.py
@@ -233,6 +233,7 @@ class NpmVersionRange(VersionRange):
         ">=": ">=",
         "<": "<",
         ">": ">",
+        "=": "=",  # This is not a native node-semver comparator, but is used in the gitlab version range for npm packages.
     }
 
     @classmethod
@@ -972,6 +973,51 @@ class OpensslVersionRange(VersionRange):
             constraint = VersionConstraint(comparator="=", version=version_obj)
             constraints.append(constraint)
         return cls(constraints=constraints)
+
+
+class GitLabVersionRange(VersionRange):
+    @classmethod
+    def from_gitlab_native(cls, scheme, string):
+        vrc = RANGE_CLASS_BY_SCHEMES[scheme]
+        constraint_items = []
+        constraints = []
+        split = " "
+        if scheme == "pypi":
+            split = ","
+        pipe_separated_constraints = string.split("||")
+        for pipe_separated_constraint in pipe_separated_constraints:
+            space_seperated_constraints = pipe_separated_constraint.split(split)
+            constraint_items.extend(space_seperated_constraints)
+        comparator = ""
+        # A constraint item can be a comparator or a version or a version with comparator
+        # If it's empty continue
+        # If it's in `vers_by_native_comparators`, append it with the comparator and continue
+        # If it's a version, make version constraint from the version and use the comparator from the previous item and make comparator empty
+        # If it's a version with comparator, use split_req to get version and comparator to form constraint and make comparator empty
+        for constraint_item in constraint_items:
+            if not constraint_item:
+                continue
+            if "".join([comparator, constraint_item]) in vrc.vers_by_native_comparators:
+                comparator = "".join([comparator, constraint_item])
+                comparator = vrc.vers_by_native_comparators[comparator]
+                continue
+            if comparator:
+                constraints.append(
+                    VersionConstraint(
+                        comparator=comparator, version=vrc.version_class(constraint_item)
+                    )
+                )
+            else:
+                comparator, version_constraint = split_req(
+                    constraint_item, vrc.vers_by_native_comparators
+                )
+                constraints.append(
+                    VersionConstraint(
+                        comparator=comparator, version=vrc.version_class(version_constraint)
+                    )
+                )
+            comparator = ""
+        return vrc(constraints=constraints)
 
 
 vers_by_github_native_comparators = {
--- a/tests/test_version_range.py
+++ b/tests/test_version_range.py
@@ -4,11 +4,13 @@
 #
 # Visit https://aboutcode.org and https://github.com/nexB/univers for support and download.
 
+import json
 from unittest import TestCase
 import pytest
 
 from univers.version_constraint import VersionConstraint
 from univers.version_range import GemVersionRange
+from univers.version_range import GitLabVersionRange
 from univers.version_range import NugetVersionRange
 from univers.version_range import InvalidVersionRange
 from univers.version_range import PypiVersionRange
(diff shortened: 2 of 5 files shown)


src/univers/version_range.py
```diff
@@ -260,6 +260,37 @@ def from_native(cls, string):
...
+    @classmethod
+    def from_gitlab_native(cls, string):
+        constraints = []
+        parsed_constraints = set()
+        pipe_separated_constraints = string.split("||")
+        for pipe_separated_constraints in pipe_separated_constraints:
+            space_seperated_constraints = pipe_separated_constraints.split(" ")
+            constraints.extend(space_seperated_constraints)
+        comparator = ""
+        while "" in constraints:
```
pombredanne, line 273: What about a bit more explicit way:
```
# remove empty strings and empty values
constraints = [c for c in constraints if c]
```

src/univers/version_range.py
```diff
@@ -260,6 +260,37 @@ def from_native(cls, string):
...
+                parsed_constraints.add(
+                    VersionConstraint(comparator=comparator, version=cls.version_class(constraint))
+                )
+            else:
+                comparator, version_constraint = split_req(
+                    constraint, cls.vers_by_native_comparators
+                )
+                parsed_constraints.add(
+                    VersionConstraint(
+                        comparator=comparator, version=cls.version_class(version_constraint)
```
pombredanne, line 289: May be a bit more explicit?
```suggestion
                        comparator="=", version=cls.version_class(version_constraint)
```

src/univers/version_range.py
```diff
@@ -583,6 +646,14 @@ def from_native(cls, string):
 
         return cls(constraints=constraints)
 
+    @classmethod
+    def from_gitlab_native(cls, string):
+        constraints = []
+        pipe_separated_constraints = string.split("||")
+        for pipe_separated_constraints in pipe_separated_constraints:
```
pombredanne, line 653: IMHO this may not work as the item and list have the same name?

src/univers/version_range.py
```diff
@@ -319,6 +351,37 @@ def from_native(cls, string):
 
         return cls(constraints=constraints)
 
+    @classmethod
+    def from_gitlab_native(cls, string):
+        constraints = []
+        parsed_constraints = []
+        pipe_separated_constraints = string.split("||")
+        for pipe_separated_constraints in pipe_separated_constraints:
```
pombredanne, line 359: IMHO this may not work as the item and list have the same name?

src/univers/version_range.py
```diff
@@ -319,6 +351,37 @@ def from_native(cls, string):
...
+            space_seperated_constraints = pipe_separated_constraints.split(" ")
+            constraints.extend(space_seperated_constraints)
+        comparator = ""
+        while "" in constraints:
```
pombredanne, line 363: May be just skip empty constraints in the loop below instead? the while/remove is a somewhat uncommon

src/univers/version_range.py
```diff
@@ -319,6 +351,37 @@ def from_native(cls, string):
...
+            constraints.remove("")
+        for constraint in constraints:
```
pombredanne, line 365: I have a hard time to understand this loop. What is the input data looking like? adding a docstring would help me

src/univers/version_range.py
```diff
@@ -260,6 +261,37 @@ def from_native(cls, string):
 
         return cls(constraints=anyof_constraints)
 
+    @classmethod
+    def from_gitlab_native(cls, string):
```
pombredanne, line 265: A lot of the code seems to be duplicated in the other "from_gitlab_native" method below?
Could there be some shared function to avoid some duplication and still keep things readable?

pombredanne (changes requested):
Thanks ... see a few comments. BTW how di you assemble the test cases? They look great!

src/univers/version_range.py
TG1999: We can not expect the comparator to be "=" every time, it's derived from the string

src/univers/version_range.py
TG1999: I have added some comments for your consideration

TG1999:
> Thanks ... see a few comments. BTW how di you assemble the test cases? They look great!

@pombredanne I ran gitlab importer in VulnerableCode and logged the versions, that we were not able to parse

tests/test_version_range.py
```diff
@@ -330,3 +344,18 @@ def test_PypiVersionRange_raises_ivr_for_unsupported_and_invalid_ranges(range, w
             assert expected in str(ivre)
     else:
         assert expected == str(PypiVersionRange.from_native(range))
+
+
+@pytest.mark.parametrize("test_case", open("./tests/data/npm_gitlab.txt").read().splitlines(False))
```
pombredanne, line 349: If you have a JSON that looks like this:
```
[
    (">=1.0.1 <2.0.1 || >= 3.0.3 < 4.0.4")), "vers:npm/>=1.0.1|<2.0.1|>=3.0.3|<4.0.4"),
    (">=1.0.1 <2.0.1 || >= 3.0.3 < 4.0.4")), "vers:npm/>=1.0.1|<2.0.1|>=3.0.3|<4.0.4"),
    ...
]

```
you could process it with this
```
@pytest.mark.parametrize(["native_gitlab_range", "expected_vers_range", json.load(open("./tests/data/npm_gitlab.json"))
```


And here is a one time thing to create them to run in Python
```
# create the initial files once:
natives = open("./tests/data/npm_gitlab.txt").read().splitlines(False))
import json
from univers.version_range import GitLabVersionRange

test_cases = []
scheme="npm"

for index, native in enumerate(natives, 1):
    vers = GitLabVersionRange.from_gitlab_native(scheme=scheme, string=native)
    test = dict(test_index=index, scheme=scheme, gitlab_native=native, expected_vers=vers)
    test_cases.append(test)

with open(f"tests/data/npm_gitlab.json", "w") as out:
    json.dump(test_cases, out, indent=2)

```
Then your test could become more or less:
```Python
@pytest.mark.parametrize("test_case", json.load(open("./tests/data/pypi_gitlab.json")))
def test_pypi_gitlab_version_range_parse(test_case):
     result = GitLabVersionRange.from_gitlab_native(
        scheme=test_case["scheme"],
        string=test_case["gitlab_native"],
     )
    assert str(result) == test_case["expected_vers"]
```

The benefit is that you are then effectively testing more.
Also the index is to help diagnostic and find which test fails

pombredanne (changes requested):
Thank you. Here are a few extra nits for your review.

src/univers/version_range.py
```diff
@@ -233,6 +233,7 @@ class NpmVersionRange(VersionRange):
         ">=": ">=",
         "<": "<",
         ">": ">",
+        "=": "=",  # This is not a native node-semver comparator, but is used in the gitlab version range for npm packages.
```
pombredanne, line 236: :+1:

tests/data/pypi_gitlab.txt
```diff
@@ -0,0 +1,122 @@
+>=1.0a1,<1.0.5||>=1.1b1,<1.1.1
```
pombredanne, line 1: I guess this and the other text files can be removed now?

pombredanne (approved):
Thanks for the updates. Just a few nits left for your consideration and then please go ahead to merge and release!

Issue #45, opened 2022-03-11
Handle possible duplication of constraints in VersionRange


keshav-space:
Eliminate the possibility of constraints duplication in VersionRange and add some tests to validate the same.

```python
>>> from univers.version_range import NginxVersionRange
>>> range==NginxVersionRange.from_native("1.5.0+, 1.4.1+, 1.4.0+")
>>> str(range)
'vers:nginx/>=1.4.0|>=1.4.1|<1.5.0|<1.5.0|>=1.5.0'
```
Context: https://github.com/nexB/univers/pull/42#discussion_r824237012

PR #69, opened 2022-05-11
Fix constraints duplication and refactor GemVersion

keshav-space:
- Refactor GemVersion
- Fix the constraint duplication in VersionRange
- Add test for constraint duplication
Closes: https://github.com/nexB/univers/issues/45


src/univers/gem.py
```diff
@@ -242,6 +237,33 @@ def __gt__(self, other):
...
+        find_segments = re.compile(r"[0-9]+|[a-z]+", re.IGNORECASE).findall
+        segments = []
+        for seg in find_segments(self.version):
+            if seg.isdigit():
+                seg = int(seg)
+            segments.append(seg)
+        return segments
+
+    @property
+    def canonical_segments(self):
```
pombredanne, line 255: a yield looks simpler:

```suggestion
    def canonical_segments(self):
        """
        Yield  "canonical segments" for this version using
        the Rubygems way for canonicalization.
        """
        for segments in self.split_segments():
            segs = dropwhile(lambda s: s == 0, reversed(segments))
            yield from sorted(segs, reverse=True)
```

src/univers/gem.py
```diff
@@ -252,40 +274,33 @@ def bump(self):
...
-                while any(isinstance(s, str) for s in segments):
-                    segments.pop()
-                segments = (str(s) for s in segments)
-                self._release = GemVersion(".".join(segments))
-            else:
-                self._release = self
-
-        return self._release
+        if self.prerelease():
+            segments = self.segments
```
pombredanne, line 298:
```suggestion
            segments = list(self.segments)
```

src/univers/nuget.py
```diff
@@ -164,6 +163,9 @@ def __lt__(self, other):
         # Revision is the same, so ignore it for comparison purposes.
         return self._base_semver < other._base_semver
 
+    def __hash__(self):
```
pombredanne, line 166: This may be best in another PR together with a test that this is indeed hashable (e.g. usable as a set value and comparable by id with "is")

pombredanne (changes requested):
Thank you . Here are a few comments for your consideration.

src/univers/nuget.py
keshav-space: Okay, will move this to another PR along with the tests.

pombredanne:
@keshav-space BTW, you do not need to wait to fix the semantic-version issue...
For reference this https://github.com/rbarrois/python-semanticversion/blob/7dcc42d2a828adbbeb6f8a23cdca40a3c61782bc/semantic_version/base.py#L412 is missing the `build` in the  
precedence key. 

Now the plot thickens, as per https://semver.org/#spec-item-10 :

> Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700, 1.0.0-beta+exp.sha.5114f85, 1.0.0+21AF26D3—-117B344092BD.

A change will therefore likely be controversial upstream in @rbarrois code.

IMHO the rationale is that we always want a stable sort order of versions and ignoring the build segment when sorting versions does not make sense as this lead to random ordering. Reporting two otherwise identical versions with different build as being the "same" or "equivalent" or "having the same precedence" but not "equal" is IMHO useful but this should not go in the way of a sensible stable sorting/ordering of  versions sequence. 

@keshav-space for now you can subclass all right to create a correct sorting

The issue:
```Python
>>> from semantic_version import *
>>> v1=Version('1.2.3+123')
>>> v1
Version('1.2.3+123')
>>> v1.precedence_key
(1, 2, 3, (MaxIdentifier(),))
>>> v2=Version('1.2.3+234')
>>> v2.precedence_key
(1, 2, 3, (MaxIdentifier(),))
>>> sorted([v2, v1])
[Version('1.2.3+234'), Version('1.2.3+123')]
>>> sorted([v1, v2])
[Version('1.2.3+123'), Version('1.2.3+234')]
```
duh!
Now a simple fix:
```Python
>>> class SortableSemverVersion(Version):
...    @property
...    def precedence_key(self):
...        return super().precedence_key + (self.build,)
... 
>>> sv1=SortableSemverVersion('1.2.3+123')
>>> sv2=SortableSemverVersion('1.2.3+234')
>>> sorted([sv1, sv2])
[SortableSemverVersion('1.2.3+123'), SortableSemverVersion('1.2.3+234')]
>>> sorted([sv2, sv1])
[SortableSemverVersion('1.2.3+123'), SortableSemverVersion('1.2.3+234')]

```

much better! 

now IMHO the fact that semver specs does not define precendence does not mean that this library should not have a stable ordering

pombredanne:
Also see https://github.com/npm/node-semver/blob/c56a701f45653940ee8536eafe43b3e46c11d6cc/functions/compare-build.js#L5 and https://github.com/npm/node-semver/blob/c56a701f45653940ee8536eafe43b3e46c11d6cc/classes/semver.js#L152
even there, we have stable ordering from what I can read.

pombredanne:
> Yeah, It makes a lot more sense to have an in-house fix for this. I don't see this making its way upstream.

I think it should as a random, non-stable sort is a terribly bad thing IMHO

rbarrois:
Hey, just chiming in since I got mentioned: I agree that a "random" sort would be an issue; and that's a use-case that should be addressed one way or another in python-semanticversion.

In order to move forward, could you please open an issue in my project, explaining which behaviour you would expect as a "stable" ordering?

Thanks!

keshav-space:
@rbarrois awesome! will raise an issue for this along with the potential patch. We can take it forward from there :)

rbarrois:
> @rbarrois awesome! will raise an issue for this along with the potential patch. We can take it forward from there :)

No problem ;) I recommend starting with an issue where we can discuss the expected API / behaviour; the implementation is easy to build once that consensus has been reached ;)

src/univers/nuget.py
```diff
@@ -164,6 +164,14 @@ def __lt__(self, other):
...
+        return hash(
+            (
+                self._base_semver.to_tuple(),
```
pombredanne, line 170: Is this containing the build segment too? Based on the issues you found, should we also subclass the "semver" library used here? Or is there an opportunity to simplify and migrate this code to us e"semantic_version" instead?

tests/test_nuget.py
```diff
@@ -76,3 +76,8 @@ def test_less(self):
         self.check_order(self.assertLess, "1.0.0-pre", "1.0.0.1-alpha")
         self.check_order(self.assertLess, "1.0.0", "1.0.0.1-alpha")
         self.check_order(self.assertLess, "0.9.9.1", "1.0.0")
+
+    def test_as_set_value(self):
+        vers1 = NugetVersion.build_value("1.0.1")
+        vers2 = NugetVersion.build_value("1.0.1")
+        assert {vers1, vers2} == {vers1}
```
pombredanne, line 83: You could test more simply with "is" to validate that the id are the same: `assert vers1 is vers2` which is the essence of a proper hash.
The set test is derived and not as explicit. You could also rename this to `test_NugetVersion_hash` .... but now you are really testing the semver library here IMHO with build_value... but what you want is really testing the NugetVersion instead or in addition.

1. add a test for the plain `semver` library to highlight the issues of not using the build segment in the hash.
2. add tests that create plain NugetVersion with build segments and various 4th segment and so on that really test the issue you found and check is on these object not on the "build_value" output.

otherwise I feel like you may not be testing what you need to test

pombredanne (changes requested):
Thanks! see some feedback inline. You have test failures too.

src/univers/nuget.py
keshav-space: Yeah, this contains build value too, that's the reason we are using to_tuple() here.

We would need some workaround to migrate this to `sementic_version` 
We would also need support for case insensitive comparison of prerelease in `sementic_version` 
for example, as of now in `sementic_version` `1.0.0-BETA` != `1.0.0-beta`

tests/test_nuget.py
keshav-space: > You could test more simply with "is" to validate that the id are the same: `assert vers1 is vers2` which is the essence of a proper hash.

I don't think `is` will work here because `is` is used to compare the identity of the object [i.e. object's memory address] and two objects with the same hash can have different identities if they reside in different memory addresses.

```python
>>> from univers.versions import NugetVersion as NV
>>> t = NV.build_value("1.2.3+12")
>>> t.__hash__()
2747090334158322414
>>> l = NV.build_value("1.2.3+12")
>>> l.__hash__()
2747090334158322414
>>> t is l
False
>>> id(t)
4517202288
>>> id(l)
4519102112
>>> k = l
>>> id(k)
4519102112
>>> k is l
True
```
Yeah, it's is bit misleading I will add a separate test for both `python-semver` and plain text `NugetVersion`

src/univers/nuget.py
pombredanne: > Yeah, this contains build value too, that's the reason we are using to_tuple() here.


So make sure you actually tests this. I do not see an explicit unit test for that below.

> We would also need support for case insensitive comparison of prerelease in sementic_version
for example, as of now in sementic_version 1.0.0-BETA != 1.0.0-beta

Then this is maybe another issue

- nuget treats release as case-insensitive https://github.com/NuGet/NuGet.Client/blob/9e244ebc11baf68ea2da0b7595b11726d0da99e5/test/NuGet.Core.Tests/NuGet.Versioning.Test/VersionComparerTests.cs#L13
- node-semver  may treat these as case sensitive.  https://github.com/npm/node-semver/blob/c56a701f45653940ee8536eafe43b3e46c11d6cc/test/fixtures/comparisons.js#L28

keshav-space: > So make sure you actually tests this. I do not see an explicit unit test for that below.

Sure, will add a test there too.

tests/test_nuget.py
pombredanne: Good point on the hash vs. id.! You could also just compare calling `assert hash(v1) == hash(v2)`

keshav-space:
Tracking for the inconsistent affected version range in `CVE-2020-5220`, causing tests to fail. https://github.com/Sylius/SyliusResourceBundle/issues/455

Context: https://github.com/nexB/vulnerablecode/wiki/WeeklyMeetings#strange-gitlab-version-ranges

keshav-space:
Something weird is happening here
```python
>>> from univers.utils import SortableSemverVersion
>>> ssv1 = SortableSemverVersion.coerce("1.3")
>>> ssv2 = SortableSemverVersion("1.3.0")
>>> ssv1 == ssv2
True
>>> ssv1 > ssv2
False
>>> ssv1 < ssv2
True
>>> ssv1
Version('1.3.0')
>>> ssv2
SortableSemverVersion('1.3.0')
```

The root of this problem
https://github.com/rbarrois/python-semanticversion/blob/c2ec5114a7865c5fccd745329ecdc34d912d9b74/semantic_version/base.py#L262

pombredanne (commented):
Is this duplicated (in part?) with #75 ?

keshav-space:
> Is this duplicated (in part?) with #75 ?

Yes, basically I have moved the two unrelated commits to separate pr #75

pombredanne:
@keshav-space gentle ping... let's find a way to get this merged!

keshav-space:
> @keshav-space gentle ping... let's find a way to get this merged!

https://github.com/nexB/univers/pull/69#issuecomment-1141400016, https://github.com/nexB/univers/pull/69#issuecomment-1156166913
At least one of these should be resolved to unblock this PR

PR #70, opened 2022-05-12
Add Nuget versions tests

pombredanne:
This PR adds several tests pulled from the NuGet client git repo and ported from C# to Python. Several of these are not yet fully working and are marked as skipped or expected failures.


tests/nuget_versioning_tests.README
```diff
@@ -0,0 +1,32 @@
+﻿![NuGet logo](https://raw.githubusercontent.com/NuGet/Home/dev/meta/resources/nuget.png)
```
TG1999, line 1: This should be saved with an ".md" extension, it can't be previewed right now

TG1999 (approved):
LGTM! A small nitpick for your consideration

tests/nuget_versioning_tests.README
pombredanne: Fixed. Thanks!

PR #72, opened 2022-05-12
Add composer and golang versions 

TG1999:
and also add support for their gitlab version ranges

### DIFF
--- a/src/univers/version_range.py
+++ b/src/univers/version_range.py
@@ -664,7 +664,16 @@ class ComposerVersionRange(VersionRange):
     # TODO composer may need its own scheme see https//github.com/nexB/univers/issues/5
     # and https//getcomposer.org/doc/articles/versions.md
     scheme = "composer"
-    version_class = versions.SemverVersion
+    version_class = versions.ComposerVersion
+
+    vers_by_native_comparators = {
+        "==": "=",
+        "<=": "<=",
+        ">=": ">=",
+        "<": "<",
+        ">": ">",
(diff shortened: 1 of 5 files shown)


src/univers/version_range.py
```diff
@@ -973,49 +991,46 @@ def from_native(cls, string):
...
-                )
-            else:
-                comparator, version_constraint = split_req(
-                    constraint_item, vrc.vers_by_native_comparators
-                )
-                constraints.append(
-                    VersionConstraint(
-                        comparator=comparator, version=vrc.version_class(version_constraint)
-                    )
+def from_gitlab_native(scheme, string):
```
pombredanne, line 994: Is the scheme exactly that of GitLab? or this is ours and there has been a mapping already done upstream in VulnerableCode?

pombredanne (approved):
LGTM! I just have one question.

src/univers/version_range.py
TG1999: This scheme is exactly from GitLab itself, the tests that I have added here are from GitLab only and not changed at all.

pombredanne: Good, so they use purl types like we do then! ... good.

pombredanne (approved):

src/univers/version_range.py
TG1999: One thing, what do you mean by scheme here? The version range or the package type itself? We map the type from slug in yaml file to package url type, and then use as scheme here

pombredanne (approved):
LGTM! :+1:

PR #75, opened 2022-06-21
Refactor gem and make nuget hashable

keshav-space:
- Refactor GemVersion
- Make NuGet version hashable
- Moved from https://github.com/nexB/univers/pull/69

### DIFF
--- a/src/univers/gem.py
+++ b/src/univers/gem.py
@@ -199,19 +199,14 @@ class GemVersion:
         if not self.is_correct(version):
             raise InvalidVersionError(version)
 
-        # If version is an empty string convert it to 0
         version = str(version).strip()
-
         self.original = version
 
+        # If version is an empty string convert it to 0
         if not version:
             version = "0"
 
         self.version = version.replace("-", ".pre.")
-        self._segments = ()
-        self._canonical_segments = ()
-        self._bump = None
-        self._release = None
 
     def __str__(self):
         return self.original
@@ -225,7 +220,7 @@ class GemVersion:
         return self.version == other.version
 
     def __hash__(self):
-        return hash(tuple(self.canonical_segments))
+        return hash(self.canonical_segments)
 
     def __eq__(self, other):
         return self.canonical_segments == other.canonical_segments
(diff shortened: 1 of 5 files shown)


src/univers/gem.py
```diff
@@ -242,6 +237,26 @@ def __gt__(self, other):
...
+            yield int(seg) if seg.isdigit() else seg
+
+    @property
+    def canonical_segments(self):
+        """
+        Yield  "canonical segments" for this version using
+        the Rubygems way for canonicalization.
+        """
+        for segments in self.split_segments():
+            segs = list(dropwhile(lambda s: s == 0, reversed(segments)))
```
pombredanne, line 257: I reckon you inherited from this but this is a rather complex expression IMHO... is there a way to rewrite this that would be easier to read?
Or at least some comment that explain what this is doing?

src/univers/gem.py
```diff
@@ -242,6 +237,26 @@ def __gt__(self, other):
     def __ge__(self, other):
         return self.__cmp__(other) >= 0
 
+    @property
+    def segments(self):
```
pombredanne, line 241: You should find a way such that segments is a list or tuple and not an iterable as this is always short and too much burden on the caller.... especially I would expect a property to return an actual collection not to yield things.
One thing could be to rename this (and the canonical one too) to `def _segments(self):` and have `a @property def segments(self): return tuple(self._segments())` ? this makes it simpler for the caller and does not break the existing "api" and behavious

src/univers/gem.py
```diff
@@ -225,10 +220,10 @@ def equal_strictly(self, other):
         return self.version == other.version
 
     def __hash__(self):
-        return hash(self.canonical_segments)
+        return hash(tuple(list(self.canonical_segments)))
```
pombredanne, line 223: Not to double call tuple + list ... but also IMHO, canonical_segments should return a tuple... see the other comment below.
```suggestion
        return hash(tuple(self.canonical_segments))
```

src/univers/gem.py
```diff
@@ -225,10 +220,10 @@ def equal_strictly(self, other):
...
 
     def __eq__(self, other):
-        return self.canonical_segments == other.canonical_segments
+        return list(self.canonical_segments) == list(other.canonical_segments)
```
pombredanne, line 226: This would not be needed with the suggestion below.

pombredanne (changes requested):
Thanks... See my review comments. Please keep `segments` and `canonical_segments` as tuples, e.g. the new property should return a tuple() not a generator.

src/univers/gem.py
keshav-space: It's an original adoption from RubyGems [`version.rb`](https://github.com/ruby/ruby/blob/13305bf0c9a03f24b5ab667784ff1dc1f10b3d9f/lib/rubygems/version.rb#L374-L379)
```rb
  def canonical_segments
    @canonical_segments ||=
      _split_segments.map! do |segments|
        segments.reverse_each.drop_while {|s| s == 0 }.reverse
      end.reduce(&:concat)
  end
```

Will add comments there and make it return a tuple instead of generator

pombredanne (approved):
LGTM. Thanks!

PR #77, opened 2022-07-22
Make tests work when not using virtualenv

sbruder:
Currently, some of them explicitly reference black via its path inside
the virtual environment. In an environment where the venv is not used,
but rather the distribution’s packages are used for making the
dependencies like black available, this does not work. This changes it
to use the black binary from the PATH.

Since the official development guidelines document activating the
virtualenv before running pytest, which will make the black binary
available, it will still work when using the virtual environment.

The CI systems (Travis and Azure Pipelines) didn’t activate the
virtualenv but rather called pytest directly from the venv. This also
changes this so CI also works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,9 @@ python:
   - "3.9"
 
 # Scripts to run at install stage
-install: ./configure --dev
+install: |
+  ./configure --dev
+  source venv/bin/activate
 
 # Scripts to run at script stage
-script: venv/bin/pytest
+script: pytest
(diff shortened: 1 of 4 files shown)


gytis-ivaskevicius (approved):

pombredanne (approved):
@sbruder Thank you ++
This makes sense alright. The activation of the virtualenv in the CI is not essential but is harmless so I am merging this now

pombredanne:
@sbruder  PS: I am always interested on how this library is used. If you care to share that would be great!

armijnhemel:
> @sbruder PS: I am always interested on how this library is used. If you care to share that would be great!

Not yet. This was packaged as part of NGI0 work. I am planning on using it in BANG.

pombredanne:
@armijnhemel Thanks!

PR #78, opened 2022-07-22
Add Nix flake as a build system

sbruder:
This PR is a part of the [Summer of Nix](https://summer.nixos.org/) project, which is funded by an EU grant connected to [NGI Zero](https://www.ngi.eu/ngi-projects/ngi-zero/) and [NLNet](https://nlnet.nl/). We are packaging software funded by and tightly related to NGI Zero projects.

It adds a build expression for [Nix](https://nixos.org/), a purely functional package manager that helps with achieving reproducibility.

Please let me know if you have any questions about this.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,25 @@
+{
+  description = "Parse and compare all the package versions and all the ranges. From debian, npm, pypi, ruby and more. Process all the version range specs and expressions.";
+
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
+  };
+
+  outputs = { self, nixpkgs }: ({
+    overlays.default = import ./overlay.nix;
+  }) // (
+    let
+      supportedSystems = [ "aarch64-darwin" "aarch64-linux" "x86_64-darwin" "x86_64-linux" ];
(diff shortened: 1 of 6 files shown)


flake.nix
```diff
@@ -0,0 +1,25 @@
...
+
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
+  };
+
+  outputs = { self, nixpkgs }: ({
+    overlays.default = import ./overlay.nix;
+  }) // (
+    let
+      supportedSystems = [ "x86_64-linux" "aarch64-linux" ];
```
pombredanne, line 12: FWIW, this runs on all OS and arches.

shell.nix
```diff
@@ -0,0 +1,3 @@
+(import (fetchTarball https://github.com/edolstra/flake-compat/archive/master.tar.gz) {
```
pombredanne, line 1: same comment as above... is the latest version the best way?

pombredanne (commented):
Thanks! See a few nits inline.

flake.nix
pombredanne: The point I was making is that the code is not architecture-specific at all, so IMHO we should not constraint supportedSystems? But I may not understand everything there.

pombredanne:
Sorry for the late merge!

PR #82, opened 2022-08-26
Fix Npm version range #79

TG1999:


TG1999:
The test data is from https://github.com/nodejs/security-wg

src/univers/version_range.py
```diff
@@ -245,35 +279,66 @@ def from_native(cls, string):
...
+                        constraints.extend(
+                            get_npm_version_constraints_from_semver_npm_spec(constraint, cls)
+                        )
+                        continue
+                    constraint = constraint.lstrip("vV")
+                    constraints.append(
+                        VersionConstraint(comparator=comparator, version=vrc(constraint))
+                    )
+                else:
+                    if constraint.endswith(".x"):
```
pombredanne, line 321: This may be simpler as a single if?
```suggestion
                    if constraint.endswith(".x") or if constraint.startswith("~") or constraint.startswith("^"):
```

src/univers/version_range.py
```diff
@@ -245,35 +279,66 @@ def from_native(cls, string):
...
+                if "".join([comparator, constraint]) in cls.vers_by_native_comparators:
+                    comparator = "".join([comparator, constraint])
+                    comparator = cls.vers_by_native_comparators[comparator]
+                    continue
+                if comparator:
+                    if constraint.endswith(".x"):
+                        constraints.extend(
+                            get_npm_version_constraints_from_semver_npm_spec(constraint, cls)
+                        )
+                        continue
```
pombredanne, line 315: would an else be easier to read? The other `continue` before seem OK, but I find it hard for me to have this `continue` and the ones below used instead of an `else` here.

src/univers/version_range.py
```diff
@@ -245,35 +279,66 @@ def from_native(cls, string):
...
+                        comparator, version_constraint = split_req(
+                            constraint, cls.vers_by_native_comparators, default="="
+                        )
+                        version_constraint = version_constraint.lstrip("vV")
+                        constraints.append(
+                            VersionConstraint(
+                                comparator=comparator, version=vrc(version_constraint)
+                            )
+                        )
+                comparator = ""
```
pombredanne, line 340: I am a bit unclear about whet you get the comparators from

pombredanne (changes requested):
Thanks!
Here are a few comments for your consideration.
I think we need a few more explicit tests that validate that a native constraints ends up parsed in a proper expected VersionRange

pombredanne (commented):
Sorry ... test-wise I missed https://github.com/nexB/univers/pull/82/files#diff-58adb926ae29bae08bfc66394189fbf5204b81a6234f6f9428d15d5f2673b2fb ...!
So ignore the test comments.

src/univers/version_range.py
TG1999: Makes sense

src/univers/version_range.py
TG1999: Please see line 307

pombredanne (approved):
LGTM. Thanks!

TG1999:
merging!

Issue #91, opened 2022-11-09
Add utility to get inverse/complement of a version range


TG1999:
There is no way currently to get an inverted version range.

TG1999:
closed by #92

PR #92, opened 2022-11-09
Add function to invert constraints and ranges #91

TG1999:

We have invert of every operator:
```
>= : <
<= : >
!= : =
< : >=
> : <=
= : !=
```
But we don't have any inversion for star operator currently, what should we do for that ?


TG1999:
@pombredanne I made inversion of each comparator here:
```
INVERTED_COMPARATORS = {
    ">=": "<",
    "<=": ">",
    "!=": "=",
    "<": ">=",
    ">": "<=",
    "=": "!=",
    "*": "^",
    "^": "*",
}
```
would like to get your opinion on this

src/univers/version_constraint.py
```diff
@@ -39,6 +39,14 @@ def operator_star(a, b):
     return True
 
 
+def operator_null(a, b):
+    """
+    Comparison operator for the star "^" constraint comparator. Since it does not matches
```
pombredanne, line 44: Is star a caret "^"?

src/univers/version_constraint.py
```diff
@@ -48,6 +56,7 @@ def operator_star(a, b):
     ">": operator.gt,
     "=": operator.eq,
     "*": operator_star,
+    "^": operator_null,
```
pombredanne, line 59: Do we really want to introduce this new syntax? Please explain the rationale and need in detail in a comment or in PR comment.

src/univers/version_constraint.py
```diff
@@ -150,10 +159,10 @@ def from_string(cls, string, version_class):
         if comparator not in COMPARATORS:
             raise ValueError(f"Unknown comparator: {comparator!r}")
 
-        if not version and comparator != "*":
+        if not version and comparator not in ["*", "^"]:
```
pombredanne, line 162: Nothing should be just nothing and not caret ^ IMHO

src/univers/version_range.py
```diff
@@ -164,6 +176,33 @@ def from_versions(cls, sequence):
             constraints.append(constraint)
         return cls(constraints=constraints)
 
+    def inverse(self):
```
pombredanne, line 179:
```suggestion
    def invert(self):
```

pombredanne (changes requested):
Thanks... see some comments for your review

tests/data/inverse-data.json
```diff
@@ -0,0 +1,5276 @@
+[
```
pombredanne, line 1: Are you sure you need all these cases? It looks like a many of them are testing identical things? Let's try to keep only one tests of each kind.

src/univers/version_range.py
```diff
@@ -164,6 +174,33 @@ def from_versions(cls, sequence):
...
+        NpmVersionRange(constraints=(VersionConstraint(comparator='<', version=SemverVersion(string='1.0.0')),))
+        """
+        inverted_constraints = []
+
+        if len(self.constraints) == 1 and self.constraints[0].comparator == "*":
+            # The inverse of "*" is an empty range.
+            return None
+
+        for constraint in self.constraints:
+            if constraint.comparator in INVERTED_COMPARATORS:
```
pombredanne, line 191: By construction it is impossible to get an unknown comparator so late n the process, so you can safely do this instead and do not need to ever raise anything
```suggestion
            inverted_comparator = INVERTED_COMPARATORS[constraint.comparator]
```

src/univers/version_range.py
```diff
@@ -164,6 +174,33 @@ def from_versions(cls, sequence):
...
+                inverted_comparator = INVERTED_COMPARATORS[constraint.comparator]
+            else:
+                raise NotImplementedError(
```
pombredanne, line 194:
```suggestion
```

src/univers/version_range.py
```diff
@@ -164,6 +174,33 @@ def from_versions(cls, sequence):
...
+                    f"Cannot invert a range with a {constraint.comparator!r} comparator."
+                )
+            inverted_constraint = VersionConstraint(
```
pombredanne, line 197: Should this rather be an invert()  method on VersionConstraint?

pombredanne (changes requested):
Thanks!
see my review inline for your consideration.
IMHO the inversion should be first a method on a constraint and then second a range method that use the constraints method.
We would need a few unit tests rather than only data driven tests, and we could get fewer tests by removing duplicated tests.

src/univers/version_constraint.py
```diff
@@ -132,6 +132,28 @@ def __lt__(self, other):
         # we compare tuples, version first
         return (self.version, self.comparator).__lt__((other.version, other.comparator))
 
+    def is_star(self):
+        return self.comparator == "*"
+
+    def invert(self):
+        INVERTED_COMPARATORS = {
```
pombredanne, line 139: Please add a docstring

tests/test_version_range.py
```diff
@@ -393,3 +393,20 @@ def test_npm_advisory_version_range_parse(test_case):
         string=test_case["npm_native"],
     )
     assert str(result) == test_case["expected_vers"]
+
+
+def test_inverse():
```
pombredanne, line 398: You need tests for the lower level constraint.invert e.g.,  for https://github.com/nexB/univers/pull/92/files#diff-c77d34029ce1991650937489359eef45029f37be4f1f27c21f56e50e64209d9aR138 too

pombredanne (changes requested):
Thanks. Here are a few extra nit pickings for your consideration.

src/univers/version_constraint.py
TG1999: Done!

pombredanne (approved):
LGTM!
:+1:

PR #99, opened 2023-02-08
Drop python 3.6 and use 3.11 in CI

TG1999:


AyanSinhaMahapatra (approved):
LGTM! Thanks!

Oops, too soon, there's more errors, maybe typing related?

.travis.yml
```diff
@@ -10,10 +10,11 @@ dist: xenial
 
 language: python
 python:
-  - "3.6"
   - "3.7"
   - "3.8"
   - "3.9"
+  - "3.10"
```
AyanSinhaMahapatra, line 16: Ubuntu 16.04 seems to be too old and doesn't have the newer python wheels?
We might want to just use azure here anyway, since we moved away from travis in other repos. See https://github.com/nexB/skeleton/commit/1165f12eb321cf3ecced9655424d6a376933c23c for reference

TG1999: Thanks for this deprecating travis.

AyanSinhaMahapatra (approved):
LGTM, Thanks!

Feel free to merge.

PR #108, opened 2023-03-14
Support Normalization of VersionRange

keshav-space:
## Why Normalize VersionRange?

A `VersionRange` represents set of segments of package versions. Different VersionRange can represent the same set of segments. 

For example, consider the package `pkg:npm/foobar` with the following versions: `["1.0", "2.1", "2.2", "3.0", "3.1", "5.0"]`.
We can represent certain segments using the `vers` expression `vers:npm/<=2.2|>=3.0|<3.1|5.0`. 
This can also be represented as:

- `vers:npm/>=1.0|<3.0|>=3.0|<3.1|5.0`
- `vers:npm/<3.1|5.0`
- `vers:npm/>=1.0|<=2.2|>=3.0|<3.1|5.0`

These different representations make it difficult to validate whether two different VersionRange represent the
same versions of a package.

To effectively compare these ranges, we need to normalize them. The idea is that all the `vers` referring to
the same set of package versions should normalize to an identical `VersionRange`.
To achieve this, we take a `vers` along with all the versions of the package and generate a new `vers`
such that it contains not arbitrary version segments but only the longest contiguous segments of versions, leading to the same VersionRange expression which is identical for a particular set of versions.

If we apply this normalization to the above example of `pkg:npm/foobar`:

- `vers:npm/<=2.2|>=3.0|<3.1|5.0` would normalize to `vers:npm/>=1.0|<=3.0|5.0`

The same normalization applies to other VersionRange since they all refer to the same set of versions,
they will all normalize to the exact same VersionRange:

- `vers:npm/>=1.0|<3.0|>=3.0|<3.1|5.0` => `vers:npm/>=1.0|<=3.0|5.0`
- `vers:npm/<3.1|5.0` => `vers:npm/>=1.0|<=3.0|5.0`
- `vers:npm/>=1.0|<=2.2|>=3.0|<3.1|5.0` => `vers:npm/>=1.0|<=3.0|5.0`
-------------------------------------
- Support normalization of vers range expression using all the available package version.
- Support VersionRange from snyk advisory and discrete version.
- Various bug fix in `from_gitlab_native` https://github.com/nexB/univers/issues/136
- Bug fix https://github.com/nexB/univers/issues/137


pombredanne (commented):
Thanks
LGTM overall! we have likely to find a bettr way than using the native intbitset of at least have a fallback to plain Python, I will review further later this WE!

keshav-space:
> I think we could:
> 
> 1. extract Span as its own mini library also reused in ScanCode? May be this is overakill though
> 2. have a fallback to a plain builtin set when intbitset is not installed... and have this also in SCTK

Added `fallback to builtin set`

src/univers/normalized_range.py
```diff
@@ -0,0 +1,329 @@
...
+            return
+
+        version_type = VERSIONS_BY_PACKAGE_TYPE.get(purl_type)
+        sorted_versions = sorted([version_type(i) for i in versions])
+        sorted_versions = [version.string for version in sorted_versions]
+        index = list(range(1, len(sorted_versions) + 1, 1))
+        return dict(zip(sorted_versions, index))
+
+    @classmethod
+    def from_github(cls, range_expression: Union[str, List], purl_type: str, all_versions: List):
```
pombredanne, line 132: This is not GitHub but a GitHub Advisory and same for Gitlab: ghsa, github_advisory gitlab_advisory

src/univers/normalized_range.py
```diff
@@ -0,0 +1,329 @@
...
+def strip_leading_v(version: str):
+    """
+    Return version without leading v.
+    """
+    if not version.startswith("v"):
+        return version
+    return version[1:]
+
+
+VERSIONS_BY_PACKAGE_TYPE = {
```
pombredanne, line 288: This does double duty with existing code

pombredanne (changes requested):
Thanks! See some/many comments for your consideration!

src/univers/version_range.py
```diff
@@ -446,7 +480,7 @@ def from_native(cls, string):
         return cls(constraints=constraints)
 
 
-def split_req(string, comparators, default=None, strip=""):
+def split_req(string, comparators, comparators_rear={}, default=None, strip=""):
```
pombredanne, line 483: what is the definition of  `comparator_rear`?

Also, if you want to account for the Maven, GitLab, Snyk and NuGet range syntaxes, you likely want to have another function for this entirely.

Unless there are cases in the wild where the two notations -- intervals boundaries with `[]()` vs. logical `><=`) -- are mixed together in a single expression?  I would surprised that this ever happens, hence two functions would be best.

Do you have some example links to paste?

tests/test_version_range.py
```diff
@@ -456,3 +456,30 @@ def test_mattermost_version_range():
             VersionConstraint(comparator=">=", version=SemverVersion("5.0")),
         ]
     ) == VersionRange.from_string("vers:mattermost/>=5.0")
+
+
+def test_version_range_normalize_case1():
+    known_versions = ["3.0.0", "1.0.0", "2.0.0", "1.3.0", "1.1.0", "1.2.0"]
```
pombredanne, line 462: What about adding an extra "4.0.0" out of the range entirely?

pombredanne (changes requested):
Thanks. See review comments.

src/univers/version_range.py
keshav-space: Done!

src/univers/version_range.py
```diff
@@ -1223,14 +1271,142 @@ def build_range_from_github_advisory_constraint(scheme: str, string: str):
...
+        delimiter = "," if "," in item else " "
+        if delimiter == ",":
+            snyk_constraints = item.strip().replace(" ", "")
+            constraints = snyk_constraints.split(",")
+        else:
+            snyk_constraints = item.strip()
+            constraints = snyk_constraints.split(" ")
+
+        for constraint in constraints:
+            if any(comp in constraint for comp in "[]()"):
```
pombredanne, line 1362: :+1:

src/univers/version_range.py
```diff
@@ -1223,14 +1271,142 @@ def build_range_from_github_advisory_constraint(scheme: str, string: str):
...
+                version_constraints.append(
+                    VersionConstraint(
+                        comparator=comparator,
+                        version=version,
+                    )
+                )
+    return vrc(constraints=version_constraints)
+
+
+def build_range_from_discrete_version_string(scheme: str, string: Union[str, List]):
```
pombredanne, line 1380: Would this be a possible dupe of https://github.com/nexB/univers/blob/8f0d7276e13686d93fee5313641b9803f4d4cb15/src/univers/version_range.py#L166 ?

pombredanne (approved):
Thank. Looking fine. There are a few nits for your review. ANd please merge then!

src/univers/version_range.py
keshav-space: Yup, droping it in favour of from_versions.

Issue #26, opened 2022-01-11
`PyPiVersionRange.from_native` not working


Hritik14:
https://github.com/nexB/univers/blob/aa2253d17a11be3af36278be37e898c05fdbc4fd/src/univers/version_range.py#L501-L504

and PEP 440 shows an example for version specifier as follows:
```
~= 0.9, >= 1.0, != 1.3.4.*, < 2.0
```
See: https://www.python.org/dev/peps/pep-0440/#version-specifiers

```py
>>> from univers.version_range import PypiVersionRange
>>> rng_from_native = PypiVersionRange.from_native("~= 0.9, >= 1.0, != 1.3.4.*,
 < 2.0")
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-20-af3412ff72fa> in <module>
----> 1 rng_from_native = PypiVersionRange.from_native("~= 0.9, >= 1.0, != 1.3.4.*, < 2.0")

~/Contrib/univers/src/univers/version_range.py in from_native(cls, string)
    514             operator = spec.operator
    515             version = spec.version
--> 516             assert isinstance(version, cls.version_class)
    517             comparator = cls.vers_by_native_comparators[operator]
    518             constraint = VersionConstraint(comparator=comparator, version=version)

AssertionError:
```

Something is not right

GermanMT:
I have programmed a small piece of code that solves the problem with the ~= operator and the versions ending in *. I have disregarded the === operator as mentioned within the TODOs of the PypiVersionRange class as PEP 440 considers it a non-recommended approach. I haven't had time to add it to a pull request as I needed to add this to a project of my own, but if needed I could do so and contribute to the project.

```python
def sanitize_pip_constraints(raw_constraints: str) -> str:
    constraints = []
    for raw_constraint in raw_constraints.split(','):
        operator, version = raw_constraint.strip().split(' ')
        if '==' in operator and '*' in version:
            pos = version.find('*')
            version = version[:pos - 1]
            constraints.append('>= ' + version)
            constraints.append('< ' + version[:pos - 2] + str(int(version[pos - 2]) + 1))
        elif '!=' in operator and '*' in version:
            pos = version.find('*')
            version = version[:pos - 1]
            constraints.append('< ' + version)
            constraints.append('>= ' + version[:pos - 2] + str(int(version[pos - 2]) + 1))
        elif '~=' in operator:
            parts = version.split('.')
            parts[-2] = str(int(parts[-2]) + 1)
[...]

PR #126, opened 2023-10-19
Fix PyPiVersionRange working with * and ~=

GermanMT:
Fix the problem with the ~= operator and the versions ending in *, exposed in the [issue](https://github.com/nexB/univers/issues/26).

### DIFF
--- a/src/univers/version_range.py
+++ b/src/univers/version_range.py
@@ -664,11 +664,6 @@ class PypiVersionRange(VersionRange):
         ">=": ">=",
         "<": "<",
         ">": ">",
-        # per https://www.python.org/dev/peps/pep-0440/#compatible-release
-        # For a given release identifier V.N, the compatible release clause is
-        # approximately equivalent to the pair of comparison clauses:
-        # >= V.N, == V.*
-        "~=": None,
         # 01.01.01 is NOT equal to 1.1.1 using === which is strict string
         # equality this is a rare and eventually non-suggested approach
         "===": None,
@@ -681,7 +676,7 @@ class PypiVersionRange(VersionRange):
         Raise an a univers.versions.InvalidVersion
         """
         # TODO: environment markers are yet supported
-        # TODO: handle  .* version, ~= and === operators
+        # TODO: handle === operators
 
         if ";" in string:
             raise InvalidVersionRange(f"Unsupported PyPI environment marker: {string!r}")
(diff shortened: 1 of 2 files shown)


src/univers/version_range.py
```diff
@@ -729,6 +722,35 @@ def from_native(cls, string):
...
+    def sanitize_constraints(cls, string):
+        constraints = []
+        try:
+            specifiers = SpecifierSet(string)
+        except InvalidSpecifier as e:
+            raise InvalidVersionRange() from e
+        for spec in specifiers:
+            operator = spec.operator
+            version = spec.version
+            if "*" in version:
```
pombredanne, line 735: Could we limit this to trailing "*" and not at any position?
I am not sure we can handle sanely other cases for now and the case with a star not at the end should likely be an exception.  
```suggestion
            if version.endswith("*"):
```

src/univers/version_range.py
```diff
@@ -729,6 +722,35 @@ def from_native(cls, string):
...
+                pos = version.find("*")
+                version = version[: pos - 1]
+                if "==" in operator:
+                    constraints.extend(
+                        [f">= {version}", f"< {version[:pos - 2]}{str(int(version[pos - 2]) + 1)}"]
+                    )
+                elif "!=" in operator:
```
pombredanne, line 742:
```suggestion
                elif operator == "!=":
```

src/univers/version_range.py
```diff
@@ -729,6 +722,35 @@ def from_native(cls, string):
...
+                    constraints.extend(
+                        [f"< {version}", f">= {version[:pos - 2]}{str(int(version[pos - 2]) + 1)}"]
```
pombredanne, line 744: Same nit as above: there must be a more readable way to express this.

src/univers/version_range.py
```diff
@@ -729,6 +722,35 @@ def from_native(cls, string):
 
         return cls(constraints=constraints)
 
+    @classmethod
+    def sanitize_constraints(cls, string):
```
pombredanne, line 726: Could you add tests specifically for this function? And some doc string?

pombredanne (changes requested):
Thank you ++! See some comments for your consideration.

pombredanne:
BTW, do you have URLs to some concrete examples of where these constraints are used in the wild? Thanks!

pombredanne:
@GermanMT too bad you closed this!

PR #131, opened 2024-03-02
Support pub package manager

shravankshenoy:
Description coming soon. Stay tuned :)

### DIFF
--- a/src/univers/versions.py
+++ b/src/univers/versions.py
@@ -390,6 +390,7 @@ class DartVersion(SemverVersion):
     def build_value(cls, string):
         return super().build_value(string)
 
+
 class GolangVersion(SemverVersion):
     @classmethod
     def build_value(cls, string):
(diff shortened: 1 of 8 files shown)


src/univers/versions.py
```diff
@@ -385,6 +385,12 @@ def build_value(cls, string):
         return super().build_value(string.lstrip("vV"))
 
 
+class DartVersion(SemverVersion):
```
TG1999, line 388:
```suggestion
class PubVersion(SemverVersion):
```

src/univers/versions.py
```diff
@@ -702,4 +712,5 @@ def bump(self, index):
     OpensslVersion,
     LegacyOpensslVersion,
     AlpineLinuxVersion,
+    DartVersion,
```
TG1999, line 715:
```suggestion
    PubVersion,
```

TG1999 (changes requested):
Minor nits for your consideration, also please remove all unrelated changes.

shravankshenoy:
> Minor nits for your consideration, also please remove all unrelated changes.

Thanks @TG1999 for the initial review, I will replace all instances of DartVersion with PubVersion. 

Regarding the other un related changes, I believe they were caused by the black formatter. I have `black=22.3.0` as mentioned in requirements-dev.txt, but it is still causing the changes. Infact when I ran `black --check .`, I got a bunch of files which seem to not follow black standards, as shown in image below

image

Could you check if you are getting a similar result when you run the command? If so I would be glad to create a separate PR which fixes all the formatting issues.

Moreover, even the tests which are failing in the Azure pipeline are due to Black formatting, although I have run black on all the files I have created in this PR. My hunch is the files listed above might be causing the problem, but let me know if that is not the case.

image

PR #146, opened 2024-11-13
Introduce "all" and "none" specifiers

immqu:
This PR introduces the possibility to specify _all_ versions, or _none_ using the notation ```vers:all/*``` and ```vers:none/*```. To allow this, it simply adds a version check in the contains method that returns true if all is specified, and false if none is specified.
Addresses the spec improvement suggested in https://github.com/package-url/purl-spec/issues/267


tschmidtb51:
Do we check somewhere that the strings are just `vers:all/*` or `vers:none/*` instead of something invalid as `vers:all/*|>17.4` or `vers:none/!4.2`?

tschmidtb51:
@immqu Please also check that your commits are signed (see DCO fails).

immqu:
> Do we check somewhere that the strings are just `vers:all/*` or `vers:none/*` instead of something invalid as `vers:all/*|>17.4` or `vers:none/!4.2`?

Currently, it is checked whether the ```*``` constraint is included, and if so, no other constraint is allowed. So, no adjustment is necessary to prevent something like ```vers:all/*|>17.4``` (see https://github.com/aboutcode-org/univers/blob/main/src/univers/version_range.py#L141). However, it is possible to specify something like ```vers:all/>1.2.3``` (without the ```*```). 
I have added a strict string equality check now, though.

tests/test_version_range.py
```diff
@@ -552,6 +552,11 @@ def test_version_range_all():
     all_vers = VersionRange.from_string("vers:all/*")
     assert all_vers.contains(Version("1.2.3"))
     assert PypiVersion("2.0.3") in all_vers
+    # test for invalid all range specification
+    with pytest.raises(Exception):
+        VersionRange.from_string("vers:all/>1.2.3")
+    with pytest.raises(Exception):
+        VersionRange.from_string("vers:all/*|>1.2.3")
```
tschmidtb51, line 559: The same should be done for `vers:none`

immqu:
See also the corresponding PR for [purl](https://github.com/package-url/purl-spec/pull/343)

PR #147, opened 2024-11-20
Lexicographic versioning scheme

immqu:
This PR introduces a simple string versioning scheme that compares strings based on lexicographical order.


immqu:
Addresses https://github.com/package-url/purl-spec/issues/265

Issue #132, opened 2024-05-24
Add support for swift vers range


keshav-space:
Swift uses semantic versioning. Refer to the official Swift documentation https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html#version

The following are the native version ranges supported in the Swift manifest and their corresponding vers expression:

| Native  | Vers |
| ------------- | ------------- |
| `exact: "1.2.3"`  | `vers:swift/1.2.3` |
| `from: "1.2.3"` | `vers:swift/>=1.2.3\|<2.0.0` |
| `"1.2.3"..<"1.2.6"` | `vers:swift/>=1.2.3\|<1.2.6` |
| `"1.2.3"..."1.2.6"` | `vers:swift/>=1.2.3\|<=1.2.6` |

https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html#package-dependency

Lokeshranjan8:
@keshav-space im little bit of confused . Here we have to create a file  test_swift_version_range.py ? right

PlayerIUnknown:
Added PR #156 to solve this issue. Urging the maintainers to take a look.

PR #156, opened 2025-03-27
Added Swift Native Version Range Support and Associated Unit Tests

PlayerIUnknown:
#### Overview
This pull request introduces support for Swift's native version ranges within the Univers project. Swift utilizes semantic versioning, and this update enables the handling of Swift package manager's specific version range declarations effectively. The update includes parsing logic for Swift version ranges and corresponding unit tests to ensure functionality adheres to Swift's versioning specifications.

#### Details
Swift's package manager allows specifying dependencies with various version range formats. This implementation supports the following native Swift version range specifications:

- **Exact Version**: Specified using `exact: "X.Y.Z"`, where the dependency must match an exact version.
- **From Version**: Specified using `from: "X.Y.Z"`, indicating a minimum version up to the next major version but not including.
- **Range with Upper Bound**:
  - **Exclusive Upper Bound**: Using `"X.Y.Z"..<"A.B.C"` to specify a range including the lower bound but excluding the upper bound.
  - **Inclusive Upper Bound**: Using `"X.Y.Z"..."A.B.C"` to specify a range including both the lower and upper bounds.

#### Modifications
- **SwiftVersionRange Class**: A new class to parse and handle Swift version ranges, integrated into the Univers versioning system.
- **Unit Tests**: Comprehensive tests are added in `tests/test_swift_versions.py` to verify parsing and handling of all supported version range specifications.

#### Example Usage
The `SwiftVersionRange` can be instantiated with a Swift version range string:
```python
vr = SwiftVersionRange('from: "1.2.3"')
```
This will parse and prepare constraints that ensure compatibility checks against provided Swift package versions.

#### Testing
Unit tests cover:
- Exact version match
- Version ranges starting from a specific version up to but not including the next major version
- Exclusive and inclusive upper bounds in version ranges

#### Impact
This update allows developers using the Univers project to integrate Swift packages more seamlessly, adhering to Swift's native versioning schemes. It enhances the versatility of the Univers version range handling and ensures better cross-platform package management compatibility.

### DIFF
--- a/src/univers/swift_version_range.py
+++ b/src/univers/swift_version_range.py
@@ -1,0 +1,57 @@
+import attr
+
+from univers.version_constraint import VersionConstraint
+from univers.version_range import VersionRange
+from univers.versions import SwiftVersion
+
+
+@attr.s(auto_attribs=True, frozen=True)
+class SwiftVersionRange(VersionRange):
+    expression: str
+    scheme: str = "swift"
+    version_class: type = SwiftVersion
+    constraints: list = attr.ib(init=False)
+
+    def __attrs_post_init__(self):
+        object.__setattr__(self, "constraints", self.parse(self.expression))
+
+    @staticmethod
+    def parse(expression):
+        constraints = []
+        # Remove quotes for parsing.
+        expression = expression.replace('"', "").strip()
+
+        # Handle different range types.
+        if "..<" in expression:
+            parts = expression.split("..<")
+            if len(parts) == 2:
+                lower = SwiftVersion(parts[0].strip())
+                upper = SwiftVersion(parts[1].strip())
+                constraints.append(VersionConstraint(comparator=">=", version=lower))
+                constraints.append(VersionConstraint(comparator="<", version=upper))
+        elif "..." in expression:
+            parts = expression.split("...")
+            if len(parts) == 2:
+                lower = SwiftVersion(parts[0].strip())
+                upper = SwiftVersion(parts[1].strip())
+                constraints.append(VersionConstraint(comparator=">=", version=lower))
+                constraints.append(VersionConstraint(comparator="<=", version=upper))
+        else:
+            # Handle other cases such as 'exact:' and 'from:' prefixes.
+            if "exact:" in expression:
+                version = SwiftVersion(expression.split("exact:")[1].strip())
+                constraints.append(VersionConstraint(comparator="=", version=version))
+            elif "from:" in expression:
+                version = SwiftVersion(expression.split("from:")[1].strip())
+                next_major_version = version.next_major()
+                constraints.append(VersionConstraint(comparator=">=", version=version))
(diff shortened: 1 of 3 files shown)


PlayerIUnknown:
This solves the issue : "Add support for swift vers range" #132. Urging maintainers to review it.

src/univers/swift_version_range.py
keshav-space, line 1: This should not be here, it belongs in `version_range.py`.

src/univers/swift_version_range.py
```diff
@@ -0,0 +1,57 @@
...
+from univers.version_constraint import VersionConstraint
+from univers.version_range import VersionRange
+from univers.versions import SwiftVersion
+
+
+@attr.s(auto_attribs=True, frozen=True)
+class SwiftVersionRange(VersionRange):
+    expression: str
+    scheme: str = "swift"
+    version_class: type = SwiftVersion
```
keshav-space, line 12: How about using `SemverVersion`, if that's what the Swift version is?

```suggestion
    version_class: type = versions.SemverVersion

```

src/univers/swift_version_range.py
```diff
@@ -0,0 +1,57 @@
...
+    constraints: list = attr.ib(init=False)
+
+    def __attrs_post_init__(self):
+        object.__setattr__(self, "constraints", self.parse(self.expression))
```
keshav-space, line 16: Why is this needed? Aren’t we already taking care of it in the base `VersionRange` class?

src/univers/swift_version_range.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @staticmethod
+    def parse(expression):
```
keshav-space, line 19: This isn’t correct, you need to implement `from_native`.

src/univers/swift_version_range.py
```diff
@@ -0,0 +1,57 @@
...
+                constraints.append(VersionConstraint(comparator="<", version=next_major_version))
+            else:
+                # Handle a single version without any prefix.
+                version = SwiftVersion(expression)
+                constraints.append(VersionConstraint(comparator="=", version=version))
+
+        return constraints
+
+    def __str__(self):
+        return f"vers:swift/{'|'.join([c.to_string() for c in self.constraints])}"
```
keshav-space, line 57: Is there a specific reason why we need to override the base __str__ for swift?

keshav-space (changes requested):
Thanks @PlayerIUnknown, see some suggestions below. Also make sure to go through the base VersionRange definition and take a look at other version range implementations like ConanVersionRange and NpmVersionRange.

tests/test_swift_versions.py
```diff
@@ -0,0 +1,38 @@
+import unittest
```
pombredanne, line 1: Do you  links to actual example of these version ranges seen in the wild?

PR #159, opened 2025-06-25
Add py.typed marker file

jaimergp:
Needed to tame CI in https://github.com/package-url/packageurl-python/pull/186


pombredanne:
Note: the CI failures are because we need to update the set of base images, and OS versions used.

PR #160, opened 2025-07-16
Add lexicographic versioning scheme

immqu:
This PR introduces a simple string versioning scheme that compares strings based on lexicographical order (UTF-8 encoded).

Addresses package-url/vers-spec#24

### DIFF
--- a/tests/test_versions.py
+++ b/tests/test_versions.py
@@ -13,6 +13,7 @@ from univers.versions import EnhancedSemanticVersion
 from univers.versions import GentooVersion
 from univers.versions import GolangVersion
 from univers.versions import IntdotVersion
+from univers.versions import LexicographicVersion
 from univers.versions import MavenVersion
 from univers.versions import NginxVersion
 from univers.versions import NugetVersion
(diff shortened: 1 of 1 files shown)


src/univers/versions.py
```diff
@@ -146,6 +146,29 @@ def is_valid(cls, string):
...
+    """
+    Create a string, even if, e.g., an integer is given
+    """
+
+    @classmethod
+    def normalize(cls, string):
+        return remove_spaces(str(string))
+
+    def __lt__(self, other):
+        return self.value.encode("utf-8") < other.value.encode("utf-8")
```
pombredanne, line 163: Is there some reference docs that explain how two UTF-8 byte strings are compared, so we can get that working beyond Python?

darakian: It looks like this conforms with rfc 3629 which states 

> UTF-8 encodes UCS characters as a varying number of octets, where the number of octets, and the value of each, depend on the integer value assigned to the character in ISO/IEC 10646 (the character number, a.k.a. code position, code point or Unicode scalar value).  This encoding form has the following characteristics (all values are in hexadecimal):
> ...
> * The byte-value lexicographic sorting order of UTF-8 strings is the same as if ordered by character numbers.  Of course this is of limited interest since a sort order based on character numbers is almost never culturally valid.

https://www.rfc-editor.org/rfc/rfc3629.txt

darakian (approved):
for clarity I approve of adding this to the spec

pombredanne (approved):
Let's merge!

PR #190, opened 2026-04-22
Raise error for deb comparators << and >>

immqu:
This PR addresses the recent fix in the vers spec which clarifies that the deb comparators << and >> are not valid vers input, see https://github.com/package-url/vers-spec/issues/51 and https://github.com/package-url/vers-spec/pull/64.
Normalization still works as expected, e.g. ```DebianVersionRange.from_native("<<1.0")``` normalizes to ```vers:deb/<1.0```, but it does not accept ```vers:deb/<<1.0``` anymore.


pombredanne:
Thanks!

Issue #196, opened 2026-04-29
Observe: Require Normalized Vers


immqu:
In the vers-spec PR https://github.com/package-url/vers-spec/pull/65, we are requiring already normalized vers inputs, so once that PR is merged, univers should be updated to throw errors when non-canonical vers is used as input.

PR #200, opened 2026-05-22
Require canonical inputs

immqu:
This PR implements the [recent change in the vers-spec](https://github.com/package-url/vers-spec/issues/50) which requires canonical inputs instead of normalizing them. 
The parser now rejects:
- ASCII whitespace in VERS strings and individual constraints
- Leading or trailing pipe separators
- Consecutive pipe separators
- Non-canonical constraint ordering
- Duplicate versions across constraints
- Invalid comparator sequences
It also adds schema-driven parse tests for canonical and non-canonical cases.


immqu:
Addresses #196

